PR #74, opened 2014-11-07
Change web server root to `cwd` by default

thedeeno:
I think breaking the 'project is responsible for its own directory' convention is a pretty significant and not necessarily required (our team doesn't need to do it, for example). I propose that if users want the current default behavior it should be configurable (so the convention breaking is obvious), but by default we stick to the cross-platform/cross-language expectation.

### DIFF
--- a/runner/config.js
+++ b/runner/config.js
@@ -34,7 +34,7 @@ function defaults() {
     // Whether the default set of local (or remote) browsers should be targeted.
     remote:      false,
     // The on-disk path where tests & static files should be served from.
-    root:        path.resolve('..'),
+    root:        path.resolve('.'),
     // The component being tested. Must be a directory under `root`.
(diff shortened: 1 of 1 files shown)


jreichenberg:
WRT the seed-element pattern for dependencies, we've had success with /not/ doing the .bowerrc magic (directory: "../").  This way, dependencies are not scattered to the wind and live inside the component during development time.  Dependency paths in the component are still coded as though it would be consumed downstream (assumes dependencies are siblings). The solution to run local docs/examples from within the component is a http server with a fallback root of 'bower_components/' to try when it would otherwise 404 looking for dependencies as siblings.  Seems much more reasonable than the .bowerrc trick.

Easy in both http-server (via the 'proxy' option) and serve-static (via an additional 'app.use()' call). See also https://github.com/expressjs/serve-static/issues/35

PR #92, opened 2014-12-03
Fix case of FakeXHR when sending a FormData payload.

garlicnation:
There's a pull request in Sinon.JS to fix this upstream, but I've got tests that won't pass without the fix. https://github.com/cjohansen/Sinon.JS/pull/618


nevir:
Let's get the upstream guy merged in; I don't want to have to maintain patches to the concatenated source

PR #103, opened 2014-12-16
Send test status (pass/fail) to SauceLabs

Klathmon:
It didn't look like there was any support for sending pass/fail information to SauceLabs, so I added it in.

Only one file changes and the changes are pretty small overall. I have been using it in my personal project, but haven't tested it too much outside of that.


pope:
+1 to this feature.

PR #111, opened 2015-01-13
Fix testImmediateAsync catch timeout

pdf:
If an exception occurs running a test via `testImmediateAsync`, it will never
return as the `err` is updated, but `testComplete` is never set. This small
patch fixes this behaviour.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -6985,6 +6985,7 @@ window.testImmediateAsync = function testImmediateAsync(name, testFn) {
     });
   } catch (error) {
     err = error;
+    testComplete = true;
   }
 };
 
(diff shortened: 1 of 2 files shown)


pdf:
I don't know what is up with my CLA, you can see here for the details: https://github.com/Polymer/core-ajax/pull/57#issuecomment-69520191

addyosmani:
@nevir for the blocks in https://github.com/pdf/web-component-tester/blob/18cb213c9e61469f308a84e3d6398d1c74ec3412/browser.js#L6981, it doesn't look like `testComplete` would get updated as mentioned in the description. Change thus LGTM, but you probably have more knowledge about whether this change is needed :)

nevir:
Awesome, thanks for this, and sorry for the delay!

PR #128, opened 2015-02-04
Hang a reference to the ChildRunner's iframe-local wct object

### LINKED ISSUE #101: ChildRunner.prepare never called
`ChildRunner.prepare` in `browser/childrunner.js` does not appear to ever be called. As a result, `this.share = wct.share` never happens. Which then results in `this.emitEvent( 'sub-suite-end', childRunner.share );` not returning any data. Dido for `sub-suite-start`.

Klathmon:
Fixes #101.

I removed the "prepare" function (it seemed overly complex for just one line, and the name didn't really make sense for where it is now), and i put the reference to the share variable in.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -700,28 +700,28 @@ Progress.prototype.draw = function(ctx){
       , y = half
       , rad = half - 1
       , fontSize = this._fontSize;
-  
+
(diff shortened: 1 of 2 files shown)


nevir:
:+1:

nevir:
`v2.2.3`

PR #173, opened 2015-05-28
Fixes #172

### LINKED ISSUE #172: Flakey results can occur due to running tests while mocha's stylesheet is loading
WCT's browser.js ensures mocha is loaded and this includes a stylesheet which is dynamically loaded. WCT may run tests while this stylesheet is loading and this can lead to flakey results in tests that query stylesheet or computed style information.

It would be better to defer starting tests until the mocha stylesheet has loaded.

sorvell:
by deferring testing until dependencies are fully loaded, specifically mocha's css file.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1334,14 +1334,15 @@
   /**
    * We have some hard dependencies on things that should be loaded via
    * `environmentScripts`, so we assert that they're present here; and do any
-   * post-facto setup.
+   * post-facto setup. This setup may be asynchronous and a `callback` can be 
+   * supplied to be informed when this is done.
    */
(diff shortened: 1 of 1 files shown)


garlicnation:
LGTM. Any opinion @nevir ?

azakus:
LGTM after moving fixes into correct files

devinivy:
Since this fix, my suites aren't running at all.  Getting an error, `Timed out loading [path to suite]`.

azakus:
This seems to be breaking a whole bunch of tests. Reverting until I can figure it out.

PR #200, opened 2015-07-29
Updates window.flush to refer to Polymer.dom.flush

mrdewitt:
In Polymer 1.0 no flush method gets called since Polymer.flush and WebComponents.flush are both undefined.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1712,8 +1712,8 @@
    * @param {function()} callback
    */
   window.flush = function flush(callback) {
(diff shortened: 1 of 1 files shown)


mrdewitt:
I work for Google, so I don't think I need to sign the CLA.

azakus:
Sorry for the delay.

I can't accept the PR as is because it modifies a generated file, but I'll merge the change into the correct source file and commit that.

azakus:
Committed as 8aeb4db. Thanks!

PR #211, opened 2015-08-31
Allow `childRunner end` events to bubble up to parent reporter

ctreatma:
This fixes an issue where `sub-suite-end` events were not always
triggered when a childRunner finished its tests, which caused
`share` values to be lost for some childRunners.


azakus:
LGTM

PR #241, opened 2015-11-03
use system bower

### LINKED ISSUE #217: Drop `bower` as a dependency and spawn it instead
Related: https://github.com/PolymerElements/polymer-starter-kit/issues/337#issuecomment-142429372

The `bower` dependency tree is massive and the user should have `bower` installed globally anyways.

azakus:
Drop bower as a requirement and use the system bower instead.
If no system bower is installed, warn the user that stuff might not work.

Fixes #217

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "express": "^4.8.5",
     "findup-sync": "^0.2.1",
     "glob": "^4.0.6",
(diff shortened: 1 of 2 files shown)


azakus:
Nope, removing bower install completely :D

PR #254, opened 2015-12-08
Fix incorrect sauce syntax

robdodson:
For sauce you need to send in more browser configuration than is currently listed in the readme


azakus:
LGTM

PR #257, opened 2015-12-11
Migrate from esperanto to rollup

azakus:
Esperanto is deprecated, rollup is the replacement


MeTaNoV:
that's sad, Esperanto is such a nice language! :D

PR #258, opened 2015-12-14
Only restore stubs if they are actually sinon proxies.

btelles:
This is is the same problem as 

https://github.com/Polymer/web-component-tester/pull/251

Just a quick fix.


cdata:
LGTM :+1: Thanks for the contribution!

azakus:
This was pushed in v4.0.2 just now :)

PR #261, opened 2015-12-16
Update version number in bower.json

btelles:
Thought bower was installing the wrong version, then I looked at the repo, and it looks like this version wasn't updated. Oops! :-P


azakus:
Fixed in 87652a4ed84b0c7cc4e8c37af065a658abe520ac, I'm using `mversion` now to keep them up to date at the same time. Thanks!

PR #264, opened 2015-12-18
Update `test-fixture` repo URL in README.md

tomalec:


zinkkrysty:
This is pointing to the element for the old version of Polymer (0.8) See #265 for the fix to the right link.

tomalec:
Thanks :) However, it seems the new one still won't work with > 0.8 templates
https://github.com/PolymerElements/test-fixture#data-bound-templates

azakus:
The repo in PolymerElements is the most up to date one, so I merged #265 instead. Thanks!

PR #269, opened 2016-01-07
Add option to a11ySuite to ignore some rules by name

azakus:


azakus:
@alice @cdata @notwaldorf PTAL

Issue #270, opened 2016-01-09
npm install fails for web-component-tester on node 0.10


TimvdLippe:
`npm install` fails on installing the dependency `test-fixture`:

```shell
npm ERR! Error: No compatible version found: test-fixture@'github:PolymerElements/test-fixture'
npm ERR! Valid install targets:
npm ERR! ["0.1.0","0.1.1","1.0.0","1.0.1","1.0.2"]
npm ERR!     at installTargetsError (/usr/share/npm/lib/cache.js:719:10)
npm ERR!     at /usr/share/npm/lib/cache.js:638:10
npm ERR!     at saved (/usr/share/npm/node_modules/npm-registry-client/lib/get.js:142:7)
npm ERR!     at /usr/lib/nodejs/graceful-fs/polyfills.js:133:7
npm ERR!     at Object.oncomplete (fs.js:107:15)
npm ERR! If you need help, you may report this log at:
npm ERR!     <http://github.com/isaacs/npm/issues>
npm ERR! or email it to:
npm ERR!     <npm-@googlegroups.com>

npm ERR! System Linux 3.13.0-68-generic
npm ERR! command "/usr/bin/nodejs" "/usr/bin/npm" "install"
npm ERR! cwd /var/www
npm ERR! node -v v0.10.25
npm ERR! npm -v 1.3.10
npm ERR! 
npm ERR! Additional logging details can be found in:
npm ERR!     /var/www/npm-debug.log
npm ERR! not ok code 0
```

TimvdLippe:
The `package.json` used is

```js
{
  "name": "planningtool",
  "private": true,
  "devDependencies": {
    "browser-sync": "^2.7.7",
    "del": "^2.0.2",
    "glob-all": "^3.0.1",
    "gulp": "^3.8.5",
    "gulp-autoprefixer": "^3.1.0",
    "gulp-cache": "^0.4.0",
    "gulp-changed": "^1.0.0",
    "gulp-html-extract": "^0.1.0",
    "gulp-if": "^2.0.0",
    "gulp-imagemin": "^2.2.1",
    "gulp-jscs": "^3.0.0",
    "gulp-jscs-stylish": "^1.1.2",
    "gulp-jshint": "^1.6.3",
    "gulp-latex": "^1.0.1",
    "gulp-load-plugins": "^1.1.0",
    "gulp-minify-css": "^1.2.1",
    "gulp-minify-html": "^1.0.2",
    "gulp-rename": "^1.2.0",
[...]

PR #274, opened 2016-01-14
Fix broken reference to test-fixture repo in package.json

geppy:
I'm on Windows 10 with npm 1.4.6, and wct won't install for me.  As best I can tell from the [npm documentation](https://docs.npmjs.com/files/package.json), the test-fixture dependency is specified incorrectly.  This PR fixes that.


TimvdLippe:
This would fix issue #270

azakus:
LGTM

PR #275, opened 2016-01-21
Add some httpbin-like functionality to web component tester

rictic:
Enables iron-ajax tests to be standalone and not hit the third party httpbin.org service.

### DIFF
--- a/runner/httpbin.js
+++ b/runner/httpbin.js
@@ -1,0 +1,93 @@
+'use strict';
+
+var express = require('express');
+var bodyParser = require('body-parser');
+var multer = require('multer');
+var portScanner = require('./port-scanner');
+
+var httpbin = express.Router();
+
+function capWords(s) {
+  return s
+      .split('-')
+      .map(function (word) {return word[0].toUpperCase() + word.slice(1)})
+      .join('-');
+}
+
+function formatRequest(req) {
+  var headers = {};
+  for (var key in req.headers) {
+    headers[capWords(key)] = req.headers[key];
+  }
+  var formatted = {
+    headers: headers,
+    url: req.originalUrl,
+    data: req.body,
+    files: req.files,
+    form: {},
+    json: {},
+  };
+  var contentType = (headers['Content-Type'] || '').toLowerCase().split(';')[0];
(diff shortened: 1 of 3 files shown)


runner/webserver.js
```diff
@@ -139,6 +140,9 @@ module.exports = function(wct) {
           log:     wct.emit.bind(wct, 'log:debug'),
         }));
 
+        app.use('/httpbin', httpbin.httpbin);
+        console.log('installed httpbin');
```
azakus: Use `wct.emit('log:debug', 'installed httpbin')`

runner/httpbin.js
```diff
@@ -0,0 +1,85 @@
...
+  var port = 7777;
+  server.listen(port);
+  server.port = port;
+  serverDestroy(server);
+  cleankill.onInterrupt(function(done) {
+    server.destroy();
+    server.on('close', done);
+  });
+
+  console.log('Server running at http://localhost:7777/');
```
azakus: Use `wct.emit('log:info', 'Server running at http://localhost:7777/')`

azakus:
LGTM with changes to use the wct logging system.

runner/httpbin.js
azakus: Optional: use `portScanner` to find another sauce supported open port (`SAUCE_PORTS`)

runner/webserver.js
rictic: Just removed the line, I was trying to debug whether this line was being reached and whether `npm link` had worked like I expected it to.

runner/httpbin.js
rictic: Used portScanner.

I think console.log is appropriate for this sort of debug server functionality, as it's going to be the only thing output, and it should always be actionable.

azakus: The only real difference is that `wct.emit('log:info')` will colorize the line in the same style as the line announcing the webserver's port. I think it's nice to keep them consistent.

rictic: Talked it out offline, it wouldn't be easy to pipe this little standalone server into the rest of the machinery.

PR #277, opened 2016-01-25
Update lodash to version 4

MartinMoizard:


motss:
:+1: for this PR

justinfagnani:
Thanks for the PR! I'm checking on our internal version of lodash...

seungha-kim:
+1

MartinMoizard:
Hi @justinfagnani, did you have time to check this pull request?

justinfagnani:
We're not at 4.0 yet internally, and updating is a bit of a pain. Is there a problem with staying at 3.x for a while?

btelles:
At least one I use requires lodash 4.0.1 or higher. It's probably a small change on their part (probably a _.concat fix), but it's a bit annoying to have to work around that.

abdonrd:
News about this?

artem-vladimirov:
Sorry, guys. Any progress with this?

justinfagnani:
We'll upgrade after the TypeScript PR is in.

justinfagnani:
cc @rictic

rictic:
We need to be compatible with lodash v3 for the time being, but I'm not against using code that works across both lodash v3 and lodash v4 here. In general, now that we require node v4, I think we should prefer to do the ES6 native thing when there is a simple ES6 option. For example, `lodash.unique(foo)` can usually be written as `Array.from(new Set(foo))` .

rictic:
Closing this PR but feel free to send a new one that updates the lodash dependency to use the range `3.7 - 4.*`

PR #294, opened 2016-02-26
fixes of replace()

mkazlauskas:
The `replace` addon can only be called once, because it works by stubbing `Polymer.Base.instanceTemplate`.

I added `replaceMany` to allow multiple replacements, which can be used like this:

```javascript
replaceMany({
  'x-foo': 'x-fake-foo',
  'y-bar': 'y-fake-bar'
});
```


mkazlauskas:
I signed it!

mkazlauskas:
Added alternate email address

justinfagnani:
Hi @mkazlauskas thanks for the PR!

I'm starting to take a look at this. Assuming this goes in, it looks like we don't have documentation on `replace` in the README. Can you add copy docs for replace and add replaceMany to the README? I'll make sure it makes it into the site docs.

justinfagnani:
Ok, first general question. It looks like the basic issue with `replace` is that it always tries to stub `instanceTemplate`, and the stub only supports one replacement. Rather than adding `replaceMany`, which also always tries to stub `instanceTemplate`, but allows for multiple replacements, why don't we fix the basic naive stubbing and allow multiple calls to `replace`?

I imagine replace would look something like this:

```javascript
extendInterfaces('replace', function(context, teardown) {
  var decorator = null;
  return function replace(oldTagName) {
    return {
      with: function(tagName) {
        if (decorator === null) {
          decorator = new TagReplacer();
          stubInstanceTemplate(decorator, teardown);
        }
        decorator.replace(oldTagName, tagName);
      }
    };
  };
});
```

What do you think?

justinfagnani:
Either the docs from the site, or the docs from the comments. I just want to make sure there's some evidence of the feature in the README :)

mkazlauskas:
I made it work using new approach and added some content to README. Let me know if you have any more neat ideas :)

mkazlauskas:
@justinfagnani I noticed console.log happening from my replaced elements. Figured out there's something else wrong with `replace()` and it was. 

Initially it was calling original `instanceTemplate()` before applying changes. It imported element that is being replaced to document and (I believe so) it wasn't garbage collected after replacing it with decorator. It's better if they're not even imported to document since Polymer elements can do all kinds of nasty stuff on their own.

I made a fix by applying decorators before calling original `instanceTemplate`.

mkazlauskas:
@justinfagnani hey, what's the status of having this PR merged?

btelles:
@justinfagnani ping

azakus:
Looks like this has been stewing for far too long. @e111077, @justinfagnani let's reviewhow #310 and this differ, and get some tests landed 🚀

justinfagnani:
Definitely on deck for tomorrow

justinfagnani:
Thanks for the PR @mkazlauskas. We went with a similar but slightly different solution for the fix.

PR #295, opened 2016-02-26
Update bower dependencies

azakus:
Sort dependencies, update chai to v3 like wct uses


azakus:
@justinfagnani PTAL

justinfagnani:
LGTM

PR #308, opened 2016-04-24
Make wct.conf.json a dotfile

tomalec:
What about making `wct.conf.json` a dotfile?

It's just a config ile, and in most of projects it's not a vital one. 

It will fall gently into bower's default ignores, and require even less boilerplate for new custom/Polymer element.

### DIFF
--- a/runner/config.js
+++ b/runner/config.js
@@ -17,7 +17,7 @@ var serveWaterfall = require('serve-waterfall');
 var paths = require('./paths');
 
 var HOME_DIR       = path.resolve(process.env.HOME || process.env.HOMEPATH || process.env.USERPROFILE);
-var JSON_MATCHER   = 'wct.conf.json';
+var JSON_MATCHER   = '.wct.conf.json';
 var CONFIG_MATCHER = 'wct.conf.*';
 var WCT_ROOT       = path.resolve(__dirname, '..');
 
@@ -91,7 +91,7 @@ function defaults() {
     //     }
     //
     registerHooks: function(wct) {},
-    // Whether `wct.conf.*` is allowed, or only `wct.conf.json`.
+    // Whether `wct.conf.*` is allowed, or only `.wct.conf.json`.
     //
     // Handy for CI suites that want to be locked down.
     enforceJsonConf: false,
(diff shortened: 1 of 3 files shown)


tomalec:
Maybe if switching to `.wct.json.conf` in all projects is too much, we could just change `var JSON_MATCHER   = '*wct.conf.json';` to support both?

justinfagnani:
I feel like I see just as many config files as non-dotfiles as dotfiles. bower.json itself isn't a dotfile, neither is package.json, mocha.opts, gulpfile.js, etc., etc. I personally don't want much hidden from me in the terminal. I kind of with .travis.yml wasn't a dotfile.

Besides bower ignoring by default, is there any big gain?

runner/config.js
```diff
@@ -17,7 +17,7 @@ var serveWaterfall = require('serve-waterfall');
 var paths = require('./paths');
 
 var HOME_DIR       = path.resolve(process.env.HOME || process.env.HOMEPATH || process.env.USERPROFILE);
-var JSON_MATCHER   = 'wct.conf.json';
+var JSON_MATCHER   = '.wct.conf.json';
```
rictic: Is this using regexp rules? Do we need to escape the dot?

ankon:
I rather have less hidden files than more. `wct.conf.js(on)` is something to be checked into the repository, and dot files are a lot easier to miss when searching for weird problems -- of which web-component-tester tends to have some :)

runner/config.js
tomalec: https://github.com/js-cli/node-findup-sync/blob/master/README.md says it uses https://github.com/jonschlinkert/micromatch so we do not need escaping neither this, nor the other dots.

tomalec:
As WCT already supports `wct.conf.js` and `wct.conf.json`, what about accepting `(.|)wct.conf.js(|on)` - version with and without dot?

FredKSchott:
We are going to merge #388 when it's ready, which will let you specify a custom file name for wct.config. Please use this if you'd like to use a dotfile for your config file.

tomalec:
Great! Thanks :)

PR #310, opened 2016-04-28
added functionality to do multiple replaces

### LINKED ISSUE #306: Cannot stub more than 1 element?
When I tried to use replace() multiple times to stub out multiple elements, sinon complains:

> Error: Attempted to wrap instanceTemplate which is already wrapped
>     Object.wrapMethod at /components/sinon/sinon-1.12.1.js:1210
>           Object.stub at /components/sinon/sinon-1.12.1.js:3007
>           Object.with at web-component-tester/browser.js:1555
> 

Is replace() designed to be used more than once? Or is it only limited to only stub out 1 element?

e111077:
Fixes #306

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1491,9 +1491,14 @@ extendInterfaces('replace', function(context, teardown) {
   return function replace(oldTagName) {
     return {
       with: function(tagName) {
+
         // Keep a reference to the original `Polymer.Base.instanceTemplate`
         // implementation for later:
         var originalInstanceTemplate = Polymer.Base.instanceTemplate;
+
+        if (Polymer.Base.instanceTemplate.isSinonProxy) {
+          Polymer.Base.instanceTemplate.restore();
+        }
 
         // Use Sinon to stub `Polymer.Base.instanceTemplate`:
         sinon.stub(Polymer.Base, 'instanceTemplate', function(template) {
--- a/browser/mocha/replace.js
+++ b/browser/mocha/replace.js
@@ -22,6 +22,10 @@ extendInterfaces('replace', function(context, teardown) {
         // implementation for later:
         var originalInstanceTemplate = Polymer.Base.instanceTemplate;
 
+        if (Polymer.Base.instanceTemplate.isSinonProxy) {
+          Polymer.Base.instanceTemplate.restore();
+        }
+
         // Use Sinon to stub `Polymer.Base.instanceTemplate`:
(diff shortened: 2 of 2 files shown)


e111077:
passes the following crude tests:

https://github.com/Polymer/web-component-tester/tree/multiple-replace-with-tests

justinfagnani:
Did you see https://github.com/Polymer/web-component-tester/pull/294 ?

e111077:
Nope. Should I close the other issue?

browser.js
```diff
@@ -1992,4 +1997,4 @@ document.addEventListener('DOMContentLoaded', function() {
 });
 
 }());
-//# sourceMappingURL=browser.js.map
\ No newline at end of file
+//# sourceMappingURL=browser.js.map
```
cdata: What's the story here? Was this automatically added by something? Is there a corresponding map that should be checked in?

cdata:
> crude tests

...the tests we deserve.

cdata:
@azakus Elliot wrote actual tests for our testing tool. That was really nice of him. Do you think you could give us guidance on how to add them as actual tests for this tool?

browser.js
e111077: Git kept saying it was missing a newline at the end of the file and it does have a newline :/

cdata: Oh, sorry, I misread the diff as adding the whole sourcemap comment. SGTM!

txsmith:
We ran into the same issue today and applied this fix locally with success. Happy to report it works!

browser/mocha/replace.js
```diff
@@ -22,6 +22,10 @@ extendInterfaces('replace', function(context, teardown) {
         // implementation for later:
         var originalInstanceTemplate = Polymer.Base.instanceTemplate;
 
+        if (Polymer.Base.instanceTemplate.isSinonProxy) {
+          Polymer.Base.instanceTemplate.restore();
```
justinfagnani: Rather than restoring each time, just stub once inside the if statement.

e111077: Further investigation shows that Sinon's API does not seem to support replacing the wrapped spy. Also looking at their GH for this function, it seems as if replacing it manually would just repeat their already-written code. I recommend that we restore if it is already wrapped in a spy. (stay the same)

WDYT?

e111077:
Continued on #316

PR #312, opened 2016-04-30
Make the "flush" function chainable via a promise-like interface

btelles:
This little patch has greatly simplified my testing over the past few weeks. Any chance it can go into the library?
The additional JSDocs should be self-explanatory.


btelles:
Gentle ping? 
This shouldn't be a breaking change either. Perhaps worthy of a patch bump, orayne a minor version bump?

rictic:
I would be in favor of actual promises here. I think something that is only slightly promise-like will result in unexpected behavior, as people expect full promise chaining from promise-like objects.

As one example of what it looks like, this doesn't propagate errors correctly or allow users to catch errors by providing a second argument to `then`.

btelles:
@rictic I tried quite a bit to jimmy-rig actual promises into this method, but ran into roadblocks. It was a long time ago, so I've forgotten what they were...something about needing to reject and catch errors in order to propagate errors to the parent test suite, or something.

I think it's a _really_ useful feature though, and several people have talked about it on stackoverflow and on the internal (Google) discussion mailing list. Any chance you could give it a quick try and see if you can implement the chaining? Sounds like you're more versed in promises than I am, and I'd love to learn how to use them more effectively.

PR #313, opened 2016-05-17
Add quiet CLI option

FredKSchott:
WCT is using its own logger, so it's going to be a little work to move off of it. Luckily theres some good stuff in here that we can bring into plylog when the time comes.

This work just adds support for the quiet option in addition to the already-supported verbose option.

/cc @justinfagnani


justinfagnani:
LGTM

PR #314, opened 2016-05-25
Convert node portion of WCT to typescript

rictic:
This is a no impact, zero-functionality-changes port. It'll be followed up by more aggressive cleanup steps, but this was easier to do as an incremental step (and also ensured that I'd actually be able to keep up with any changes that may happen on master while this PR is reviewed).

I followed the conventions in https://github.com/Polymer/tools-common where possible (e.g. not changing the directory structure yet, leaving jshint for js hinting for now, as this PR can only get so large). I also didn't have luck getting noLib compilation working.


custom_typings/promisify-node.d.ts
```diff
@@ -0,0 +1,17 @@
+declare module 'promisify-node' {
+  interface NodeCallback<T> {
+    (err: any, value: T):void;
+  }
+  import {Promise} from 'es6-promise';
```
justinfagnani: Shouldn't we be getting this from `lib.core.es6.d.ts`?

custom_typings/findup-sync.d.ts
```diff
@@ -0,0 +1,13 @@
+declare module 'findup-sync' {
```
justinfagnani: Looks like we might want to configure tslint to run on `custom_typings/`

gulpfile.js
```diff
@@ -7,26 +7,81 @@
...
  */
-var fs = require('fs');
-var gulp = require('gulp');
-var jshint = require('gulp-jshint');
-var path = require('path');
-var lazypipe = require('lazypipe');
-var mocha = require('gulp-mocha');
-var rollup = require('rollup');
-var runSequence = require('run-sequence');
+'use strict';
```
justinfagnani: Did you make any changes from what's in tools-common? Let's keep track of that and upstream. Very soon I want to make that an library that we can import rather than copy and paste.

package.json
```diff
@@ -41,8 +41,13 @@
...
   ],
   "scripts": {
+    "lint": "gulp lint",
+    "init": "gulp init",
+    "build": "gulp build",
     "test": "gulp test",
-    "build": "gulp build"
+    "prepublish": "gulp build-all",
+
+    "test:watch": "watch 'gulp test:unit' runner/ browser/ bin/ test/ tasks/"
```
justinfagnani: is `watch` a global npm package? If so, can it be installed locally instead?

runner/browserrunner.ts
```diff
@@ -0,0 +1,211 @@
+/**
```
justinfagnani: Arrrgh... I wish we could see a diff here. I wonder if two commits would have worked: 1) rename to .ts 2) do the actual changes.

runner/browserrunner.ts
```diff
@@ -0,0 +1,211 @@
...
+import {Config} from './config';
+
+export interface Stats {
+  status: string;
+  passing?: number;
+  pending?: number;
+  failing?: number;
+}
+
+interface NodeCB<T> {
```
justinfagnani: Is there not a common declaration for this in node.d.ts? node-promisify has it too.

runner/browserrunner.ts
```diff
@@ -0,0 +1,211 @@
...
+// Browser abstraction, responsible for spinning up a browser instance via wd.js and
+// executing runner.html test files passed in options.files
+export class BrowserRunner {
+  timeout: number;
+  browser: wd.Browser;
+  stats: Stats;
+  sessionId: string;
+  timeoutId: NodeJS.Timer;
+
+  constructor(public emitter: NodeJS.EventEmitter, public def: BrowserDef, public options: Config, public doneCallback: NodeCB<BrowserRunner>) {
```
justinfagnani: long line

justinfagnani: I assume that adding `public` to the parameter declares it as a class field. That's... neat :) but I wouldn't mix with the long hand class field declarations.

runner/browserrunner.ts
```diff
@@ -0,0 +1,211 @@
...
+  _init(error: any, sessionId: string) {
+    if (!this.browser) return; // When interrupted.
+    if (error) {
+      // TODO(nevir): BEGIN TEMPORARY CHECK. https://github.com/Polymer/web-component-tester/issues/51
+      if (this.def.browserName === 'safari' && error.data) {
+        // debugger;
+        try {
+          const data = JSON.parse(error.data);
+          console.log(data.value.message);
+          if (data.value && data.value.message && /Failed to connect to SafariDriver/i.test(data.value.message)) {
```
justinfagnani: long line (and I know this one isn't yours!)

runner/browserrunner.ts
```diff
@@ -0,0 +1,211 @@
...
+
+  quit() {
+    this.done('quit was called');
+  }
+
+  //HACK
+  static BrowserRunner = BrowserRunner;
+}
+
+module.exports = BrowserRunner;
```
justinfagnani: how hard would it be to make this an ES-compliant module?

runner/cli.ts
```diff
@@ -7,18 +7,19 @@
  * Code distributed by Google as part of the polymer project is also
  * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
  */
-var _      = require('lodash');
-var chalk  = require('chalk');
-var events = require('events');
+import * as _ from 'lodash';
```
justinfagnani: style: we've been sorting by the import string, seems the most stable.

runner/cli.ts
```diff
@@ -30,21 +31,21 @@ var updateNotifier;
   }
 })();
 
-function run(env, args, output, callback) {
-  var done = wrapCallback(output, callback);
+export function run(env: void, args: string[], output: NodeJS.WritableStream, callback: (err: any) => void) {
```
justinfagnani: long line

justinfagnani: what's up with `env: void`?

runner/cli.ts
```diff
@@ -57,26 +58,26 @@ function run(env, args, output, callback) {
 // Note that we're cheating horribly here. Ideally all of this logic is within
 // wct-sauce. The trouble is that we also want WCT's configuration lookup logic,
 // and that's not (yet) cleanly exposed.
-function runSauceTunnel(env, args, output, callback) {
-  var done = wrapCallback(output, callback);
+export function runSauceTunnel(env: void, args: string[], output: NodeJS.WritableStream, callback: (err: any) => void) {
```
justinfagnani: long line

justinfagnani: `env: void` again

runner/cli.ts
```diff
@@ -57,26 +58,26 @@ function run(env, args, output, callback) {
...
-    var wctSauce = require('wct-sauce');
+    const wctSauce = require('wct-sauce');
     wctSauce.expandOptions(options);
 
-    var emitter = new events.EventEmitter();
-    new CliReporter(emitter, output, {});
-    wctSauce.startTunnel(options, emitter, function(error, tunnelId) {
+    const emitter = new events.EventEmitter();
+    new CliReporter(emitter, output, <config.Config>{});
+    wctSauce.startTunnel(options, emitter, function(error: any, tunnelId: string) {
```
justinfagnani: can error not be typed as `Error`?

runner/cli.ts
```diff
@@ -57,26 +58,26 @@ function run(env, args, output, callback) {
...
 // and that's not (yet) cleanly exposed.
-function runSauceTunnel(env, args, output, callback) {
-  var done = wrapCallback(output, callback);
+export function runSauceTunnel(env: void, args: string[], output: NodeJS.WritableStream, callback: (err: any) => void) {
+  const done = wrapCallback(output, callback);
 
-  var diskOptions = config.fromDisk();
-  var baseOptions = diskOptions.plugins && diskOptions.plugins.sauce || diskOptions.sauce || {};
+  const diskOptions = config.fromDisk();
+  const baseOptions: config.Config = diskOptions.plugins && diskOptions.plugins['sauce'] || diskOptions.sauce || {};
```
justinfagnani: long line, and might be clearer as a ternary

runner/cli.ts
```diff
@@ -87,10 +88,10 @@ function runSauceTunnel(env, args, output, callback) {
   });
 }
 
-function wrapCallback(output, done) {
-  return function(error) {
+function wrapCallback(output: NodeJS.WritableStream, done: (err: any) => void) {
+  return function(error: any) {
```
justinfagnani: `any` -> `Error`?

runner/clireporter.ts
```diff
@@ -0,0 +1,304 @@
...
+  locationStrip: [
+    /^https?:\/\/[^\/]+/,
+    /\?[\d\.]+$/,
+  ],
+  unimportantLocation: [
+    /^\/web-component-tester\//,
+  ]
+};
+
+type State = 'passing' | 'pending' | 'failing' | 'unknown' | 'error';
```
justinfagnani: 😄

runner/clireporter.ts
```diff
@@ -0,0 +1,304 @@
...
+
+const SHORT = {
+  'internet explorer': 'IE',
+};
+
+const BROWSER_PAD = 24;
+const STATUS_PAD  = 38;
+
+interface TestEndData {
+  state: 'passing'|'failing'|'pending'|'unknown';
```
justinfagnani: `state: State`?

runner/clireporter.ts
```diff
@@ -0,0 +1,304 @@
...
+  test: string[];
```
justinfagnani: if you can add any comments... I'm wondering why `test` is a `string[]`, what's in it?

runner/clireporter.ts
```diff
@@ -0,0 +1,304 @@
...
+export class CliReporter {
+  prettyBrowsers: {[id: number]: string} = {};
+  browserStats: {[id: number]: Stats} = {};
+
+  /**
+   * The number of lines written the last time writeLines was called.
+   */
+  private linesWritten: number;
+
+  constructor(public emitter: events.EventEmitter, public stream: NodeJS.WritableStream, public options: config.Config) {
```
justinfagnani: long line

runner/clireporter.ts
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    return chalk.blue(parts.join(' '));
+  };
+
+  // General Output Formatting
+
+  log(...values: any[]): void;
+  log() {
+    let values = Array.prototype.slice.call(arguments);
```
justinfagnani: Array.from(arguments)

runner/clireporter.ts
```diff
@@ -0,0 +1,304 @@
...
+  let currLength = chalk.stripColor(str).length;
+  while (currLength < length) {
+    currLength = currLength + 1;
+    str = str + ' ';
+  }
+  return str;
+}
+
+
+module.exports = CliReporter;
```
justinfagnani: convert to JS module?

runner/config.ts
```diff
@@ -7,22 +7,79 @@
...
+import * as serveWaterfall from 'serve-waterfall';
+
+import {Capabilities} from 'wd';
+import * as paths from './paths';
+import {Plugin} from './plugin';
+import {Context} from './context';
+import {BrowserDef} from './browserrunner';
+
+
+const HOME_DIR       = path.resolve(process.env.HOME || process.env.HOMEPATH || process.env.USERPROFILE);
```
justinfagnani: `alignment ==        'bad'`

runner/context.ts
```diff
@@ -0,0 +1,191 @@
...
+import * as socketIO from 'socket.io';
+import * as http from 'http';
+import * as promisify from 'promisify-node';
+
+import * as config from './config';
+import {Plugin} from './plugin';
+import {BrowserRunner} from './browserrunner';
+
+
+interface Options {
```
justinfagnani: empty interface?

runner/context.ts
```diff
@@ -0,0 +1,191 @@
...
+ * The complete set of contributors may be found at http://polymer.github.io/CONTRIBUTORS.txt
+ * Code distributed by Google as part of the polymer project is also
+ * subject to an additional IP rights grant found at http://polymer.github.io/PATENTS.txt
+ */
+import * as _ from 'lodash';
+import * as events from 'events';
+import * as util from 'util';
+import * as socketIO from 'socket.io';
+import * as http from 'http';
+import * as promisify from 'promisify-node';
```
justinfagnani: promisify is unused?

runner/paths.ts
```diff
@@ -0,0 +1,81 @@
...
+ * directory contains an `index.html`, it is expanded to that. Otheriwse, the
+ * it expands into its children (recursively).
+ *
+ * @param {string} baseDir
+ * @param {!Array<string>} patterns
+ */
+async function expandDirectories(baseDir: string, paths: string[]): Promise<string[]> {
+  const listsOfPaths: string[][] = [];
+  for (const aPath of paths) {
+    const paths = await Promise.resolve().then(() => {
```
justinfagnani: I don't think you need the `Promise.resolve().then(...`, just `paths = await promisify(fs.stat)(...).

Also, as long as you're going to use `await`, use it more:

```
let stat = await promisify(fs.stat)(path.resolve(baseDir, aPath));
let paths;
if (!stat.isDirectory()) {
  paths = [aPath];
} else {
  let files = await promisify(fs.readdir)(path.resolve(baseDir, aPath));
  // ...
}

File: runner/steps.ts [hunk 26]
```diff
@@ -7,39 +7,62 @@
...
-var async    = require('async');
-var http     = require('http');
-var socketIO = require('socket.io');
-
-var BrowserRunner = require('./browserrunner');
-var config        = require('./config');
+import * as _ from 'lodash';
+import * as http from 'http';
+import * as socketIO from 'socket.io';
+import * as promisify from 'promisify-node';
```
justinfagnani: promisify is unused

custom_typings/promisify-node.d.ts
rictic: Done.

custom_typings/findup-sync.d.ts
rictic: Done, but I had to enable internal modules because non-es6 exports.

package.json
rictic: It is installed locally, it's a devDependency of wct now

runner/browserrunner.ts
rictic: In hindsight that would have been a good idea. I just spent a while trying to rebase on top of an initial change like that and.. git got pretty unhappy.

gulpfile.js
rictic: PR at https://github.com/Polymer/tools-common/pull/3

runner/cli.ts
rictic: Sorted all imports. Annoyingly, I couldn't find a sort-imports lint checker that does what we want here.

runner/browserrunner.ts
rictic: Easy, but it would be a breaking change for anyone who imported any of the non-ES-compliant modules.

I'm currently using a midway hack everywhere here by exposing the default value as a property on itself. That way both we support both the new and the old patterns.

I was thinking we'd do one back-compat release with typescript and then a breaking release with a bunch of stuff improved (like callbacks -> promises for example).

runner/browserrunner.ts
rictic: There isn't.

runner/cli.ts
rictic: It's not used anywhere. Changed to `any`, as that's the type of `process.env` which is how it's called.

runner/cli.ts
rictic: Done

runner/cli.ts
rictic: Not in a node callback, the only contract is that the operation was successful if `error` is `null`. AFAIK `error` can properly be anything.

runner/cli.ts
rictic: Wrapped and clarified.

runner/cli.ts
rictic: My vote is `any` unless we have positive knowledge otherwise. Better broad but accurate than narrow and false.

runner/clireporter.ts
rictic: According to the source that passes this message, it looks like `pending` isn't an option.

runner/clireporter.ts
rictic: Looks like it's the names of the tests that ran. Added comment.

runner/clireporter.ts
rictic: Ah, that's nicer.

runner/clireporter.ts
rictic: Same as above, IMO we should support both modes for the next release.

runner/config.ts
rictic: Unaligned and wrapped.

runner/context.ts
rictic: Oops, this was supplanted by Config. Removed.

runner/context.ts
rictic: Huh, surprised we didn't have a warning for that. Removed.

runner/paths.ts
rictic: Yeah, looks like I started down this road but got distracted at a working intermediate form.

Finished awaitifying.

runner/steps.ts
rictic: Removed.

runner/browserrunner.ts
rictic: Wrapped all long lines and moved all constructor arg property declarations into regular property declarations.

runner/browserrunner.ts
rictic: Done

rictic:
All comments addressed, PTAL!

justinfagnani:
LGTM!

justinfagnani:
Oh, you need to port the last commit: 7a5c477098aee46156b627592cafdf6217bd6d7a

bedeoverend:
Out of interest @rictic @justinfagnani why the decision to move WCT over to Typescript for its internals?

rictic:
@bedeoverend Yeah we're looking to move all of our node based tools to typescript. We evaluated it based on a number of factors, which I'm happy to go into, but the tl;dr is that we expect typescript to make it easier for us to maintain these tools while being invisible to existing users.

This PR is a good example. It turns out there are a bunch of kinda hairy and very underdocumented data types that made it hard for new contributors to get up to speed with WCT. Typescript lets us add that documentation while also providing tools to automatically validate it.

bedeoverend:
@rictic interesting - thanks!

fearphage:
I think this move makes it more difficult (additional hurdles) for the external JS devs to contribute. TS is pretty niche in terms of user base compared to JS. 

I understand the internal benefits, but writing in TS is just like writing in CoffeeScript for many intents and purposes. It's a common meme for people to want to contribute to a JS project, open it in Github, find out it's not written in JS, and turn around.

justinfagnani:
@fearphage: Thanks for you input. We've found that TypeScript actually increases the ability for new people to contribute because the code tends to be so much better documented wrt types. 

TypeScript and CoffeeScript couldn't be more different in this aspect, as TypeScript is just an extension of JavaScript syntax, and easily readable by anyone who knows JavaScript. TypeScript also discourages patterns that are very difficult to reason about, because they often happen to also be difficult to type check. CoffeeScript on the other had completely changes the syntax, so that it's very difficult to read by those who don't know CoffeeScript, and offers additional assistance to developers or tools to help understand the structure of the program.

We have a lot of projects and are constantly umping around between them. The ability for us to quickly understand the types that the programs deal in is critical for that. The type-checking compiler increases our velocity, and confidence in tests.

To alleviate any getting started pain, we're trying to standardize on a gulpfile and dev dependencies that means there's nothing more to do than `npm i && npm run build` to get started - no global installs are required.

fearphage:
Thank you for the quick response! 

Is JavaScript valid TypeScript? If the answer is no, then I and others now have to learn TS to contribute to this project. That's the hurdle I was referring to. That's the similarity it shares with CoffeeScript.

justinfagnani:
Yes, JavaScript is valid TypeScript.

We require type annotations where types can't be inferred in our code base (the --noImplicitAny flag) though.

fearphage:
Valid but not acceptable is better than invalid. Thank you for your replies.

rictic:
I'm also willing to take pure JS contributions, or contributions that don't typecheck, and handle getting them to typecheck. It's almost always very little work, a tiny fraction of the work necessary to write the code. Don't let the unfamiliarity of typescript prevent you from contributing!

PR #316, opened 2016-06-10
Multiple replace with tests

### LINKED ISSUE #306: Cannot stub more than 1 element?

e111077:
Branched off of #310. Fixes #306.


browser/mocha/replace.js
```diff
@@ -22,6 +22,10 @@ extendInterfaces('replace', function(context, teardown) {
         // implementation for later:
         var originalInstanceTemplate = Polymer.Base.instanceTemplate;
 
+        if (Polymer.Base.instanceTemplate.isSinonProxy) {
```
justinfagnani: Is the change we discussed yesterday still the plan here?

e111077: I added a comment in the last PR before closing it:

Further investigation shows that Sinon's API does not seem to support replacing the wrapped spy. Also looking at their GH for this function, it seems as if replacing it manually would just repeat their already-written code. I recommend that we restore if it is already wrapped in a spy. (stay the same)

WDYT?

justinfagnani: So I'm not suggesting that the wrapper is replaced, I'm suggesting that we only stub `instanceTemplate` once, with a stub that's actually capable of replacing multiple elements. This way we don't have a stub calling a stub calling a stub, each doing a querySelector and replace. It would have to be slightly rewritten to do a tree walk rather than querySelector, but that should be relatively easy.

justinfagnani: This is what I think the stub should look like:

```javascript
var replacements = {};

sinon.stub(Polymer.Base, 'instanceTemplate', function(template) {
  var dom = originalInstanceTemplate.apply(this, arguments);
  var nodeIterator = document.createNodeIterator(dom, NodeFilter.SHOW_ELEMENT);
  var node;

  while (node = nodeIterator.nextNode()) {
    var tagName = node.tagName.toLowercase();
    if (Object.hasOwnProperty(replacements, tagName)) {
      var replacement = document.createElement(replacements[tagName]);
      node.parentNode.replaceChild(replacement, node);
    }
  }
  return dom;
});
```

Besides stubbing `instanceTemplate` only on the first call, `with` would write into `replacements`.

browser/mocha/replace.js
```diff
@@ -18,6 +21,17 @@ extendInterfaces('replace', function(context, teardown) {
...
+        replacements[oldTagName] = tagName;
+
+        // If the function is already a stub, restore it to original
+        if (Polymer.Base.instanceTemplate.isSinonProxy) {
+          Polymer.Base.instanceTemplate.restore();
+        }
+
         // Keep a reference to the original `Polymer.Base.instanceTemplate`
         // implementation for later:
         var originalInstanceTemplate = Polymer.Base.instanceTemplate;
```
e111077: @justinfagnani note: this is now after the restore and thus will never be a stub.

justinfagnani: I don't think you ever need to restore though...

e111077: good point; it's all pointers at this point!

fixed

browser/mocha/replace.js
```diff
@@ -18,6 +21,17 @@ extendInterfaces('replace', function(context, teardown) {
   return function replace(oldTagName) {
     return {
       with: function(tagName) {
+        // Standardizes our replacements map
+        oldTagName = oldTagName.toLowerCase();
+        tagName = tagName.toLowerCase();
+
+        replacements[oldTagName] = tagName;
+
+        // If the function is already a stub, restore it to original
```
justinfagnani: update comment

browser/mocha/replace.js
```diff
@@ -27,27 +41,34 @@ extendInterfaces('replace', function(context, teardown) {
           // The DOM to replace is the result of calling the "original"
           // `instanceTemplate` implementation:
           var dom = originalInstanceTemplate.apply(this, arguments);
+          // creates an iterator of the node tree
```
justinfagnani: This comment doesn't add much information to the line of code. The only wood not int he code is "tree" :P

browser/mocha/replace.js
```diff
@@ -27,27 +41,34 @@ extendInterfaces('replace', function(context, teardown) {
...
-          var nodes = Array.prototype.slice.call(dom.querySelectorAll(oldTagName));
+          // Traverses the tree. A recently-replaced node will be put next, so
+          // if a node is replaced, it will be checked if it needs to be
+          // replaced again.
+          while (node = nodeIterator.nextNode()) {
+            var currentTagName = node.tagName.toLowerCase();
 
-          // For all of the nodes we want to place...
-          nodes.forEach(function(node) {
+            // If this node is in our replacements map...
```
justinfagnani: remove comment

justinfagnani:
LGTM after the last little nits. Thanks!

hungyao:
This does not seem to solve the issue observed in https://github.com/Polymer/web-component-tester/pull/294#issuecomment-191286622

replaced() elements are still being imported into the document (and causing unwanted sideeffects).

PR #322, opened 2016-06-15
replaces before creating the elements

### LINKED ISSUE #309: Replace helper does not work on Templatizer elements
The Mocha `replace` helper works by intercepting calls to `instanceTemplate` and manipulating the templates of Polymer elements before they get stamped. However, this doesn't work for custom template elements because they have a different code path for stamping themselves. The same interception needs to take place for the `templatize` method of custom template elements.

e111077:
I also built the project (forgot to do that after the last time) and added tests

Additionally fixes #309


e111077:
@justinfagnani this should do the trick, but lmk what you think about using `template._content`. It seems a bit iffy to me.

e111077:
@justinfagnani All that I changed was the last else in the traversal where we traversed up the tree. There is a while loop to traverse up until we can move laterally again. It also checks whether or not we traversed up to the top.

justinfagnani:
LGTM!

PR #326, opened 2016-06-19
Convert all callbacks to promises.

rictic:
- got rid of the async library from the node side entirely!
- must not be merged before https://github.com/Polymer/wct-local/pull/25 has landed and been released
- tried to keep prevent breaking plugins by maintaining callback compatibility with hook methods


rictic:
(the tests fail because https://github.com/Polymer/wct-local/pull/25 hasn't been merged and released yet)

justinfagnani:
Don't you still need a rebase here?

rictic:
I uh. I may have merged instead of rebasing.

The tests are still failing, but it appears to be due to a change in the test environment. I reran the tests in travis on master and it went from passing to failing with no change in wct code. It appears that chrome is no longer installed in the test environment.

I think this change is ready to review.

justinfagnani:
So should I be seeing all the convert to typescript commits or not?

rictic:
You should be seeing the commits, but not the changes.

runner/browserrunner.ts
```diff
@@ -203,7 +203,7 @@ export class BrowserRunner {
 
     // Nothing to quit.
     if (!this.sessionId) {
-      return this.doneCallback(error, this);
+      error ? this._reject(error) : this._resolve();
```
justinfagnani: standalone ternary is weird

runner/cli.ts
```diff
@@ -32,11 +32,12 @@ let updateNotifier = noopNotifier;
   }
 })();
 
-export function run(
-      env: any, args: string[], output: NodeJS.WritableStream,
-      callback: (err: any) => void) {
-  const done = wrapCallback(output, callback);
+export async function run(
+      env: any, args: string[], output: NodeJS.WritableStream) {
+  await wrap(output, _run(args, output));
```
justinfagnani: return type?

rictic: Added here and elsewhere in this PR.

runner/browserrunner.ts
rictic: This is true. Wanted: a convenient, legible, and common way of converting a node callback into a promise. This ternary saves 5 lines, and its intent seems clear to me but I'm probably too close to it to call it.

rictic:
Rebased it all clean like off of master.

runner/config.ts
```diff
@@ -521,41 +512,35 @@ function expandDeprecated(context: Context, done: (err?: any) => void): void {
...
 /**
  * @param {!Object} options The configuration to validate.
- * @param {function(*)} Callback indicating whether the configuration is valid.
  */
-export function validate(options: Config, done: (err?: string) => void): void {
+export async function validate(options: Config): Promise<void> {
   if (options['webRunner']) {
-    return done(
+    throw (
         'webRunner is no longer a supported configuration option. ' +
```
justinfagnani: can we throw only Errors?

runner/context.ts
```diff
@@ -18,12 +18,17 @@ import {BrowserRunner} from './browserrunner';
 import * as config from './config';
 import {Plugin} from './plugin';
 
-type Handler = (o: {}, callback: (err: any) => void) => void;
+type Handler = (o: {}) => Promise<any>;
```
justinfagnani: what's the point of the empty anonymous interface vs `any`, or a named interface?

runner/context.ts
```diff
@@ -18,12 +18,17 @@ import {BrowserRunner} from './browserrunner';
...
 
 /**
  * Exposes the current state of a WCT run, and emits events/hooks for anyone
  * downstream to listen to.
  *
+ * TODO(rictic): break back-compat with plugins by moving hooks entirely away
```
justinfagnani: Do there exist any plugs that aren't ours?

runner/context.ts
```diff
@@ -89,74 +94,69 @@ export class Context extends events.EventEmitter {
...
+    done = done || ((e) => {});
 
     // We execute the handlers _sequentially_. This may be slower, but it gives us
     // a lighter cognitive load and more obvious logs.
     let promise = Promise.resolve(null);
     for (const hook of boundHooks) {
       promise = promise.then(() => {
         return new Promise((resolve, reject) => {
-          hook((err) => {
+          const maybePromise = hook((err) => {
```
justinfagnani: since you're in an async function, rather than chain the promises, you can just await them.

runner/plugin.ts
```diff
@@ -43,40 +43,38 @@ export class Plugin {
   /**
    * @param {!Context} context The context that this plugin should be evaluated
    *     within.
-   * @param {function(*)} done
    */
-  execute(context: Context, done: (message?: string) => void): void {
+  async execute(context: Context): Promise<void> {
```
justinfagnani: why's this async?

runner/plugin.ts
```diff
@@ -43,40 +43,38 @@ export class Plugin {
...
   };
 
   /**
    * Retrieves a plugin by shorthand or module name (loading it as necessary).
    *
    * @param {string} name
-   * @param {function(*, Plugin)} done
    */
-  static get(name: string, done: (err: any, plugin?: Plugin) => void): void {
+  static async get(name: string): Promise<Plugin> {
```
justinfagnani: this too

runner/steps.ts
```diff
@@ -127,8 +106,7 @@ export function cancelTests(context: Context) {
 
 // Helpers
 
-function runBrowsers(
-      context: Context, done: (err?: any) => void): BrowserRunner[] {
+function runBrowsers(context: Context) {
```
justinfagnani: where'd the return type go?

runner/test.ts
```diff
@@ -60,10 +59,8 @@ import * as steps from './steps';
  *
  * @param {!Config|!Context} options The configuration or an already formed
  *     `Context` object.
- * @param {function(*)} done callback indicating error or success.
  */
-export function test(
-      options: (Config|Context), done: (err: any) => void): Context {
+export async function test(options: (Config|Context)): Promise<void> {
```
justinfagnani: style: parens around union type isn't necessary, space around `|`

runner/webserver.ts
```diff
@@ -63,10 +63,10 @@ const SAUCE_PORTS = [
  * It provides a static HTTP server for serving the desired tests and WCT's
  * `browser.js`/`environment.js`.
  */
-export function webserver(wct: Context) {
+export function webserver(wct: Context): void {
```
justinfagnani: Doesn't it seem like it'd be a little nicer if plugins were just classes (a plug-in module exported a constructor) and all the hooks were just optional methods on the class that WCT called into?

runner/context.ts
rictic: Good question, `any` is better here. I started with `{}` in a number of places because it would catch and flag assignments so I could track down all properties. This just passes the object through and we don't know anything about it (because we dispatch many messages) so `any` is better.

runner/context.ts
rictic: There do! https://www.npmjs.com/search?q=wct and https://www.npmjs.com/search?q=web-component-tester show a bunch. Google search ranking suggests that this code coverage plugin is the most popular: https://www.npmjs.com/package/web-component-tester-istanbul

runner/context.ts
rictic: Done, also did a bit more cleanup here.

runner/plugin.ts
rictic: I suppose to preserve our ability to make execute asynchronous in the future?

runner/plugin.ts
rictic: Loading a plugin eventually becoming an asynchronous action seems even more likely to me.

runner/steps.ts
rictic: It became more complicated because completion is now separate, so I'm letting type inference handle things here.

runner/test.ts
rictic: Done

runner/webserver.ts
rictic: That'd be pretty nice yeah. I'm in favor of that for v5 of wct.

rictic:
All comments addressed, PTAL

runner/plugin.ts
justinfagnani: Then do you want to await the `plugin(...)` call?

justinfagnani:
LGTM

PR #327, opened 2016-06-20
Convert unit tests to Typescript

rictic:
- depends on https://github.com/Polymer/web-component-tester/pull/326


test/integration/browser.ts
```diff
@@ -0,0 +1,583 @@
...
+/** Asserts that all browsers emitted the given errors. */
+function assertTestErrors(context: TestContext, expected: TestErrorExpectation) {
+  lodash.each(context.testErrors, function(actual, browser) {
+    expect(Object.keys(expected)).to.have.members(
+        Object.keys(actual), 'Test file mismatch for ' + browser);
+
+    lodash.each(actual, function(errors, file) {
+      const expectedErrors = expected[file];
+      // Currently very dumb for simplicity: We don't support suites.
+      expect(Object.keys(expectedErrors)).to.have.members(Object.keys(errors), 'Test failure mismatch for ' + file + ' on ' + browser);
```
justinfagnani: template strings are nice

justinfagnani:
LGTM. Please rebase/squash

rictic:
An update for those following along at home. 

On travis the first couple of integration tests (that actually launch browsers) pass, but something gets probabilistically messed up and we lose the ability to talk to them. It shouldn't be memory use because the browsers report that they have successfully closed, but I'm not sure what else it's likely to be. Removing the intervening java selenium server may make it easier to debug. 

I'm apartment hunting today but one way forward I intend to try is to patch in my changes that talk the selenium protocol directly from node itself to see if it either gets the tests to pass or at least gives us more info on what is failing and why.

rictic:
It also may make sense to remove the integration tests from the travis run temporarily and still land this change, as it's still an improvement (at least the integration tests now will run locally).

PR #329, opened 2016-06-23
fixed issue with content in light dom exceptioning

e111077:
Having a content tag in the light dom of a custom element would make the content tag lose context of its parent.


browser/mocha/replace.js
```diff
@@ -76,7 +76,11 @@ extendInterfaces('replace', function(context, teardown) {
             }
 
             if (instanceParent) {
-              instanceParent.appendChild(instanceNode);
+              Polymer.dom(instanceParent).appendChild(instanceNode);
+
+              if (instanceNode.tagName == 'CONTENT') {
```
justinfagnani: I would not appendChild twice for content tags, use an if/else

e111077:
@justinfagnani Implemented your traversal simplification, added comments to appendChild; passes tests nicely :ok_hand:

justinfagnani:
Awesome! LGTM

PR #334, opened 2016-06-27
Load the test suite on window.load

fejesjoco:


fejesjoco:
I signed it!

rictic:
Do not edit `browser.js` directly, it's a generated file. Please edit the underlying files in the `browser/` directory.

Please also include a test that only passes with the new behavior.

FredKSchott:
@fejesjoco Bump! :)

fejesjoco:
Sorry, I forgot about this.

I tried to write a test case just now but couldn't reproduce it anymore. For some reason, DOMContentLoaded was fired **after** the test script has been loaded, even if it was in ˙<head>` and not `<body>`. This shouldn't be like that, the point of DOMContentLoaded is that it is fired **before** everything has been downloaded, just after the DOM has been parsed. I guess this is a brittle assumption, it may change from browser version to browser version, and maybe even using document.write() in WCT alters this as well.

Does WCT start running the test suite as soon as DOMContentLoaded is detected at the code I was trying to change? If that's the case, it might still make sense to make this change to be sure. It should be at least guaranteed that window.load is not fired before DOMContentLoaded, so it won't cause a regression. If it occurs again that something causes a delay in window.load, WCT would wait for that. I can only hope that WCT is not doing anything when it's waiting for an event at that point, and changing what event we wait for doesn't break anything.

Otherwise, I don't know WCT that deeply, I know that a loading process can be tricky and I don't know why the code is as it is now. I can't show any strong reasons for changing it currently.

FredKSchott:
Not a problem, I'm still digging deep into wct for the first time myself.

@rictic can you weigh in here?

rictic:
Ok, it seems like `load` is better than `DOMContentLoaded` since we do use iframes. Please move these changes into the appropriate browser code rather than editing browser.js directly.

fejesjoco:
Sorry I forgot about this. I made the change as you requested.

Note that I can't reproduce the original problem anymore. I haven't even tested this yet because I don't know how to run the test suite, but will try to figure out.

fejesjoco:
I actually hit a bug that is probably caused by this. However, this PR doesn't fix it. At least now I have something to verify against. Will investigate and update this.

fejesjoco:
Changing this code only seems to be breaking things but doesn't help. The real problem seems to be elsewhere, probably around whenFrameworksReady. I'll investigate more and file a separate issue. Closing this.

fejesjoco:
There isn't even a problem in whenFrameworksReady apparently. It's a bit complicated but looks correct.

The key to the problem is that this code (and HTMLImports specifically) waits for the "interactive" readystate, not "complete". As I just found out, this state already guarantees that scripts, even remote ones, have already executed. So if someone wants to make sure that external scripts are loaded, then looking for the "interactive" readystate is perfect. This also coincides with DOMContentLoaded. (My code was buggy because it was waiting for "complete", and then it surprised me that WCT started running earlier than I expected.)

If someone has code that waits for window.load, then it can be surprising that WCT starts running sooner. Ie. it won't wait for images and iframes to load.

PR #336, opened 2016-06-29
Add an npmignore file so that npm publish will work

rictic:


justinfagnani:
What's going on with the tests? chrome is not found?

justinfagnani:
LGTM

rictic:
Yeah. AFAICT something changed in the test environment. Not sure what, but retesting passing commits shows them failing now, so it doesn't appear to be an issue with wct.

PR #343, opened 2016-07-11
Changed to simpler algorithm implementation

### LINKED ISSUE #331: Content selectors overridden when using replace
In WCT 4.3 `replace` breaks a content selector. A minimal reproducible case with shady dom:

```html
<dom-module id="my-element">
    <template>
      <other-element></other-element>
      <content-element>
        <div class="selector">
          Asdf
        </div>
      </content-element>
    </template>
  </dom-module>

  <dom-module id="other-element">
    <template>
      Other
    </template>
  </dom-module>

  <dom-module id="mock-element">
    <template>
      Mock
    </template>
  </dom-module>

  <dom-module id="content-element">
    <template>
      <content select=".selector"></content>
    </template>
  </dom-module>

  <test-fixture id="basic">
    <template>
      <my-element></my-element>
    </template>
  </test-fixture>

  <script>
    suite('Wrong replace tests', function() {
      var element;

      Polymer({is: 'my-element'});
      Polymer({is: 'other-element'});
      Polymer({is: 'mock-element'});
      Polymer({is: 'content-element'});

      setup(function() {
        replace('other-element').with('mock-element');
        element = fixture('basic');
      });

      test('does not replace correctly', function() {
        console.log(element);
      });

    });
  </script>
```

## Expected Behavior ##
Output on the screen (place a debug point on the `console.log`) should be "Mock Asdf"

## Actual Behavior ##
Output on the screen is only "Mock"

This is happening because the element fixture is replacing the other element with the mock element. However, consequently it is also removing the content selector. Then the content-element does not contain the content anymore and breaks the selector.

e111077:
Fixes #331

~~The copied template implementation caused various issues, particularly with our shady dom implementation; this commit uses the traditional polymer workflow more and relies less on the specifics of polymer the implementation and thus should be more-reliable in core polymer updates. It is also less-complex and has a lot-fewer edge cases. Steve gave a thumbs up to reusing the template.~~


e111077:
Changed to older traversal algorithm because it was simpler, has less edge cases, and relies a lot less on this specific version of Polymer's implementation.

The difference is that we clone the template itself rather than the content. This creates a template in an inert document that does not upgrade the elements inside. I then do the modifications on the inert template clone rather than on the original template.

browser/mocha/replace.js
```diff
@@ -32,87 +32,51 @@ extendInterfaces('replace', function(context, teardown) {
...
-                } else {
-                  instanceParent = instanceParent.parentNode;
-                }
+          var origContent = template._content || template.content;
+          var templateClone = document.createElement('template');
+          var content = templateClone.content;
+          var inertDoc = content.ownerDocument;
+
+          // appends children to templateClone
+          for (var n=origContent.firstChild; n; n=n.nextSibling) {
```
justinfagnani: can't you importNode and append `origContent` itself?

browser/mocha/replace.js
```diff
@@ -32,87 +32,51 @@ extendInterfaces('replace', function(context, teardown) {
...
+            // imports node from inertDoc which holds inert nodes.
+            templateClone.content.appendChild(inertDoc.importNode(n, true));
+          }
 
-                templateNode = templateNode.parentNode;
+          var nodeIterator = document.createNodeIterator(content,
```
justinfagnani: Are we testing this on IE? I saw some things about createNodeIterator (or maybe createTreeWalker) being weird on IE...

browser/mocha/replace.js
```diff
@@ -32,87 +32,51 @@ extendInterfaces('replace', function(context, teardown) {
...
-                }
+          var origContent = template._content || template.content;
+          var templateClone = document.createElement('template');
+          var content = templateClone.content;
+          var inertDoc = content.ownerDocument;
+
+          // appends children to templateClone
+          for (var n=origContent.firstChild; n; n=n.nextSibling) {
+            // imports node from inertDoc which holds inert nodes.
+            templateClone.content.appendChild(inertDoc.importNode(n, true));
```
justinfagnani: Do we have a test that tests replace with a template with bindings and/or if or repeat? @kevinpschaaf was saying we would need to copy `_notes` properties into the cloned template nodes.

test/fixtures/integration/multiple-replace/tests.html
```diff
@@ -158,6 +165,48 @@
       assert.equal(insideElement.nodeName, 'ORIGINAL-TEMPLATE-SUCCESS');
     });
   });
+
+  describe('testing template bindings', function() {
```
e111077: I added a test that checks if the bindings still work after the replacement. @kevinpschaaf or @sorvell do you forsee any issues that would arise regarding template bindings and instancing a copied template without certain binding identifiers?

43081j:
Any news on this @e111077?

The majority of my tests fail because of this and I can't really rollback to a previous version due to lack of multiple replacements. Seems to be a very critical bug as anything which makes use of `<content>` becomes untestable.

e111077:
@43081j I agree that this is a very-critical bug; this just needs one more a simple look-over by someone on the core team. I'll give you another ping when this is merged. This PR should provide a fix to the content tag issues and still support multiple-replace.

43081j:
Ah no worries, was just checking it hadn't been forgotten about.

Looking forward to the merge :)

e111077:
@justinfagnani the annotations are passed correctly.

That bug that we discussed apparently was on the single-replace master so this PR is still backwards compatible. I'll file a bug for replacing the bound value's parent in another issue.

justinfagnani:
LGTM

e111077:
@43081j Merged

PR #344, opened 2016-07-12
Fix wct local stubs

rictic:
Depends on https://github.com/Polymer/wct-local/pull/32 having been merged and released for travis to actually become green though.


justinfagnani:
LGTM

Issue #350, opened 2016-07-14
If setup() throws, the test still passes with great success


fejesjoco:
Consider this very simple test:

```
<!doctype html>
<html>
  <head>
    <meta charset="utf-8">
    <script src="web-component-tester/browser.js"></script>
  </head>
  <body>
    <script>
      suite('the-suite', () => {
        setup(() => {
          console.log('setup running');
          //throw 'oops';
        });

        test('test1', () => {
          console.log('test1 running');
        });
      });
    </script>
  </body>
</html>
```

Running this gives:

```
joco@joco:~/test$ wct test.html 
Installing and starting Selenium server for local browsers
Selenium server running on port 58264
Web server running on port 2000 and serving from /usr/local/google/home/joco/test
chrome 51                Beginning tests via http://localhost:2000/components/test/generated-index.html?cli_browser_id=0
chrome 51                Tests passed
Test run ended with great success

chrome 51 (1/0/0)                     
```

Now if you uncomment that throw sentence, it gives this:

```
joco@joco:~/test$ wct test.html 
Installing and starting Selenium server for local browsers
Selenium server running on port 58461
Web server running on port 2000 and serving from /usr/local/google/home/joco/test
chrome 51                Beginning tests via http://localhost:2000/components/test/generated-index.html?cli_browser_id=0
chrome 51                Tests passed
Test run ended with great success

chrome 51 ()                          
```

Test success/failure numbers are not shown anymore, but the final result is still success, which is very very bad. It should fail both because the suite is apparently empty and also because something threw an error.

dgl:
I came across this too and it caused much head scratching...

ppalladino:
Me too! Any ideas??

ppalladino:
My example: 
```js
suite('some-test', function() {
  let item;
     
  setup(function(done){
     // Never call done(), should timeout then throw
     // done();
  });

  test('item defined', function() {
    assert.isDefined(item, 'item defined');
  });
});
```

 outputs:

```sh
$ wct

...

$ firefox 46               Tests passed
$ Test run ended with great success
$ firefox 46 ()                         
```

Pilatch:
I discovered one way this could be addressed. Take a look at the event listeners set in runner/clireporter.js. 

In the case where the setup function fails for a test suite, that emitter object will emit events like so: sub-suite-start, test-start, sub-suite-end, sub-suite-end...

[...]

PR #351, opened 2016-07-18
Ensure fail events outside of the scope of a test result in a failure.

dgl:
Without this the failures would be visible in the HTML UI, but running
on the command line wouldn't print any failures.

Note using err.uncaught and test.type == 'hook' comes from mocha, see https://github.com/mochajs/mocha/commit/34abc7180019640dd5735be4fa2906e6410e434e for a related fix. This fixes issue #350.

### DIFF
--- a/browser.js
+++ b/browser.js
@@ -1718,6 +1718,12 @@ CLISocket.prototype.observe = function observe(runner) {
     });
   }.bind(this));
 
+  runner.on('fail', function(test, err) {
+    if (err.uncaught || test.type == 'hook') {
+      runner.emit('test end', test);
+    }
+  }.bind(this));
+
   runner.on('childRunner start', function(childRunner) {
     this.emitEvent('sub-suite-start', childRunner.share);
   }.bind(this));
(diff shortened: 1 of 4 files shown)


dgl:
I hope I've got a test for this but I'm not able to run the tests on master (even before my patch). Running `gulp test test:integration` I get an error for each test:

    1) Local Browsers hybrid "before all" hook:
       TypeError: emitter.on is not a function
        at Context.<anonymous> (test/integration/browser.js:483:15)

Pilatch:
@dgl You need to put your change to browser.js in browser/clisocket.js so that it persists between builds!

ppalladino:
👍 Looking forward to this!!

rictic:
What @Pilatch said, browser.js is a generated file that shouldn't be edited directly.

FredKSchott:
@dgl bump! :)

fejesjoco:
https://github.com/Polymer/web-component-tester/pull/415 fixed the issue.

justinfagnani:
Ok, closing this one then.

PR #361, opened 2016-08-03
removed multiple teardown calls in fixture and replace

### LINKED ISSUE #360: `Replace` teardown issue
In trying to understand more about what is happening I have added some console.log statements to the `browser.js` file around the `replace` interface extension.

Specifically I added one in the teardown  callback function.

It is called once on my first test, and then increasingly one more time per test as each test is run.  So the second test it is called two times, the third test three times and so on.

This increasing number of calls does not seem to be effective by the `describe` blocks that I traverse.  Each `it` test run just increases the number by one.

e111077:
Fixes #360. These teardowns were being attached every time the `fixture` and the `replace` functions were called, thus they were called every time they were attached at the end of each test.


rictic:
LGTM

PR #370, opened 2016-08-23
Format all code with clang-format

rictic:
Otherwise a noop change


rictic:
Justin has given me blank check verbal LGTM.

PR #372, opened 2016-08-24
Make bin/wct work with new and old versions of wct

justinfagnani:


rictic:
Looks good. Need to do the same in wct-st

rictic:
LGTM

PR #374, opened 2016-08-24
Implement testing variant component directories

rictic:
Apologies for the length of this PR, but I had to significantly rework the integration tests to test this, as well as for the sanity of the reader.
- every directory in `test/fixtures/integration/` is now a normal-looking project that you might run `wct` in
  - the expected results for each fixture is now stored in a `golden.json` file in each integration directory
- moved a few directories out of `test/fixtures/integration` that were not actually integration test related
- hooked up a couple of significant integration tests in `test/fixtures/integration` that weren't actually tested anywhere
- deleted/renamed a couple of properties options that were actually internal values that were stuck on the `config`object. 
  - I don't believe anyone is messing with these options, nor would they have much success if they did as there are many hardcoded assumptions about these values. 
  - removing our internal business from the `config` object would be the biggest thing I would change in a major
- some of the implicitly imported libraries like underscore, async, etc may or may not work in variant tests. if so, this is a feature, not a bug, as people can migrate off of depending on these implicitly included libraries as they also test their components across a broader array of dependency constraints

Naming advice please, is `variant` the right word here? However let's not block on bike shedding as the term is internal, so if possible let's try to land this ASAP so that we unblock the elements team.


justinfagnani:
Can you briefly describe where the top-level change is for testing against variant component directories?

justinfagnani:
re the "variants" name: Why not just componentDirs? The assumption can be that there's no "the" component dir, we just support 0 or more components dirs (btw, including node_modules in the future cc @FredKSchott )

rictic:
Added documentation to README.md, PTAL

rictic:
IMO componentDirs seemed a bit fuzzy for me because we already support serving files out of `node_modules`, and in the future we'll probably want to extend variants such that the `foo` variant serves out both `bower_components-foo` and `node_modules-foo` (or whatever).

justinfagnani:
We support serving _internal_ files out of node_modules, no? If a component wants to `../foo` into `node_modules/foo`, but `../bar` into `bower_components/bar` we don't support that do we?

My point is that we (probably) only support one component dir at a time from the POV of a component, so componentDirs can be a list of those.

rictic:
Places to look in the implementation:

* in config.ts we discover directories in the root dir whose names start with `bower_components-` and modify some private state in the `config` object indicating which variants were detected and the paths to the generated index files.
* in webserver.ts we set up serving so that `/components/` serves out of `bower_components`, and `/components-foo/` serves out of `bower_components-foo`
* in steps.ts we launch one browser for each variant, pointing to the generated-index files

rictic:
I'm not sure that that's the case. Line 192 of config.ts seems to just serve things indiscriminately out of `node_modules`

rictic:
(these comments should probably be on some (potentially arbitrary) part of the code so that we get threading)

rictic:
Closing in favor of #424

PR #376, opened 2016-08-25
test() function must reject when any test fails.

### LINKED ISSUE #375: Wrong exit code returned
Hi, I'm still debugging this issue, but it seems like the `wct` cli command is returning an error code `0` instead of `1` when a test fails.

I've added a few console.logs:
```
Test run ended in failure: 1 failed tests
hook: cleanup

firefox 45 (4/0/1)

process.exit 0
```

The process.exit(0) is called in https://github.com/Polymer/web-component-tester/blob/master/bin/wct#L21

I'm using Node 6.4.0, npm 3.10.3 and running in a Docker container created by image specified here: https://github.com/MartinsThiago/polymer-tester
(Note that I'm currently just running /bin/bash and running the tests that way to investigate the issue).

I can also reproduce this on Mac OS X El Capitan and the same Node/npm versions.

I'm using web-component-tester@4.3.3, and the issue persists with 4.4.0-prerelease.1.

rictic:
Fixes #375 

In the future we should have an integration test that actually shells out to `wct` and we should assert on the exit code. The integration test cleanup needs to land first though in #374

### DIFF
--- a/runner/steps.ts
+++ b/runner/steps.ts
@@ -162,7 +162,7 @@ function runBrowsers(context: Context) {
       await context.emitHook('cleanup');
 
       if (error) {
-        throw error;
+        throw new Error(error);
       }
     }())
   };
--- a/runner/steps.ts
+++ b/runner/steps.ts
@@ -161,7 +161,9 @@ function runBrowsers(context: Context) {
       // TODO(nevir): Better rationalize run-end and hook.
       await context.emitHook('cleanup');
 
-      return error;
+      if (error) {
+        throw error;
+      }
     }())
   };
 }
--- a/test/integration/browser.ts
+++ b/test/integration/browser.ts
@@ -444,7 +444,7 @@ function runsIntegrationSuite(
       try {
         await test(context);
       } catch (error) {
-        testContext.testRunnerError = error;
+        testContext.testRunnerError = error.message;
       }
     });
 
--- a/test/integration/browser.ts
+++ b/test/integration/browser.ts
@@ -44,6 +44,7 @@ function assertPassed(context: TestContext) {
 
 function assertFailed(context: TestContext, expectedError: string) {
   expect(context.runError).to.eq(expectedError);
+  expect(context.testRunnerError).to.be.eq(expectedError);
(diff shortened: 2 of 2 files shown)


runner/steps.ts
```diff
@@ -161,7 +161,9 @@ function runBrowsers(context: Context) {
       // TODO(nevir): Better rationalize run-end and hook.
       await context.emitHook('cleanup');
 
-      return error;
+      if (error) {
+        throw error;
```
FredKSchott: Are we okay throwing strings here? Can we throw an Error object instead?

rictic: Is throwing an Error more useful here? It won't have a particularly insightful stack trace.

FredKSchott: I've always strongly believed in only using Errors as exceptions as a best practice. They're much more powerful representations, and I believe most people assume that thrown/propagated errors are Error objects and not strings so you end up causing less surprises and cascading bugs in error handling.

If `runTests()` is ever exposed or called in a different way in the future, then I could definitely see this causing problems.

rictic: Ok, but in our own code we must not assume that thrown objects / rejection values are Errors. It would be nice if that constraint were enforced, but it isn't.

rictic:
Comment addressed, PTAL

justinfagnani:
LGTM

runner/steps.ts
FredKSchott: agreed, it's completely crazy that *anything* can be an exception.

FredKSchott:
LGTM!

runner/steps.ts
justinfagnani: (Weird, github wasn't showing me this conversation.)

Ideally we should follow the robustness principle wrt errors. _We_ should only throw Error objects, and we should try to handle any objects that are thrown. The problem is that it's pretty damn tedious to handle non-Error errors. Want to check for `.message` or `.stack` every time before using them? Ugh. Rethrowing is another ugly case. We don't want to wrap every rethrow.

So generally I just assume Error objects until proven otherwise. If we have plugins throwing strings, then we can deal.

PR #380, opened 2016-09-02
reverted boolean blocking multiple calls to fixture teardown

### LINKED ISSUE #360: `Replace` teardown issue

e111077:
That boolean was added to fix #360. The similar change made to the `replace` function is just fine, but the `teardown` in fixture was always being called multiple times and was not related to calling `replace`.

By adding this boolean, `<test-fixture>.restore()` is not being called which causes issues with certain elements like `app-location` since there are a million of them on the page since they are not being restored.

The work inside that function is O(1) (assuming `getElementById` is O(1)) since calling `restore` again on a fixture that has already been restored simply returns.


cdata:
LGTM 👍 

It's worth noting that @e111077 uncovered a subtle bug here (albeit one that seems benign in practice), where `teardown` will be called way more times than it needs to be to in many usages. @e111077 would you mind filing an issue about that?

PR #388, opened 2016-09-16
Added option to specify a wct config file via the commandline.

hvdb:
--configFile _name_
If specified that file will be used by WCT.
This is handy if you have multiple config files.
For instance: local and CI. Or local and against sauce/selenium grid.


hvdb:
I signed it!

README.md
```diff
@@ -184,8 +184,9 @@ you based on the suites you ask it to load._
 # Configuration
 
 The `wct` command line tool will pick up custom configuration from a
-`wct.conf.json` file located in the root of your project. It should export the
-custom configuration:
+`wct.conf.json` file located in the root of your project.  
+Or you can specify your own file via the commandline option: --configFile my.wct.conf.js
```
FredKSchott: nit: can you include the backticks to format: `--configFile my.wct.conf.js` 

Also, would prefer wording this as an example:

> Or, you can specify your own file via the `--configFile` command line option. Example: `--configFile my.wct.conf.js`

runner/config.ts
```diff
@@ -328,7 +333,7 @@ function loadProjectFile(file: string) {
  * Runs a simplified options parse over the command line arguments, extracting
  * any values that are necessary for a full parse.
  *
- * At the moment, the only values extracted are `--plugin` and `--simpleOutput`.
+ * See const: PREPARSE_ARGS for the values that are extracted.
```
FredKSchott: 👍

runner/config.ts
```diff
@@ -285,11 +290,11 @@ interface PreparsedArgs {
  * @param {string} root
  * @return {!Object} The merged configuration.
  */
-export function fromDisk(jsonOnly?: boolean, root?: string): Config {
+export function fromDisk(jsonOnly?: boolean, root?: string, configFile?: string): Config {
   const matcher = jsonOnly ? JSON_MATCHER : CONFIG_MATCHER;
 
   const globalFile = path.join(HOME_DIR, matcher);
-  const projectFile = findup(matcher, {nocase: true, cwd: root});
+  const projectFile = findup(configFile || matcher, {nocase: true, cwd: root});
```
FredKSchott: Can we handle this in runner/context.ts instead? Instead of overloading fromDisk, lets change it's function signature to: 

```js
fromDisk(matcher: string, root?: string): Config {
```

and then in context.ts:


```js
let matcher;
if (configFile) {
  matcher = configFile;
} else if (jsonOnly) {
  matcher = JSON_MATCHER;
} else {
  matcher = CONFIG_MATCHER;
}
```

hvdb:
I will have a look tomorrow!
Thx for the review

FredKSchott:
@hvdb New changes look great, but build is failing. Run `npm run build` to see & fix those errors locally

hvdb:
Hum i will look into it.
Locally it passes all tests, which is weird because i do see the error locally now.

hvdb:
@FredKSchott, so the problem is the SauceTunnel.

https://github.com/Polymer/web-component-tester/blob/master/runner/cli.ts#L74
Is uses the config.fromDisk(), that now needs a matcher because of the change done in context.ts.

So i think we need to get the options and parse the args?
Something like:
https://github.com/Polymer/web-component-tester/blob/master/runner/cli.ts#L47

Or do you see another option?

hvdb:
So i now parse options from the cmd.
and get the config from disk via the context.
https://github.com/Polymer/web-component-tester/pull/388/commits/361dd36fcb0aa58bb31bc67c7c1c1a1be5426fcd

All tests pass, but i don't use sauce.
So someone which use sauce should test if that still works....

ishmael-readingplus:
Any update on this, I'm at point where this would be helpful.

fghilardi:
This would be immensely helpful.

hvdb:
@FredKSchott Can we get some love for this pr 💘

bwolff:
+1

Indeed, it would be awesome to have this feature! We need to use different configs (local browsers) for developer and ci environments. Is there maybe a simple workaround without changing the wct.conf.json file on one of the environments?

runner/cli.ts
```diff
@@ -71,7 +71,10 @@ export async function runSauceTunnel(
     }
 
 async function _runSauceTunnel(args: string[], output: NodeJS.WritableStream) {
-  const diskOptions = config.fromDisk();
+  let cmdOptions = <config.Config>config.preparseArgs(args);
```
justinfagnani: let -> const

justinfagnani: Prefer `as` for casts.

runner/config.ts
```diff
@@ -27,6 +27,7 @@ import {Plugin} from './plugin';
 
 const HOME_DIR = path.resolve(
     process.env.HOME || process.env.HOMEPATH || process.env.USERPROFILE);
+const WCT_ROOT = path.resolve(__dirname, '..');
```
justinfagnani: is this used?

hvdb: euh nope

PR #391, opened 2016-09-23
feat(gulp): adds compatibility with Gulp 4.

### LINKED ISSUE #384: web-component-tester Gulp tasks seems not work with Gulp4
We are using web-component-tester for running tests in our polymer project, which was generated by polymer custom-build template (it has Gulp4).
But when we try to use WCT gulp task in the project as mentioned in Readme,
```
var gulp = require('gulp');
require('web-component-tester').gulp.init(gulp);

```
We are getting following error,

> assert.js:90
>   throw new assert.AssertionError({
>   ^
> AssertionError: Task function must be specified

Tried with WCT versions 4.3.1 and 4.3.4
Is this because they are not ported to Gulp4?

joaovieira:
With backward compatibility with Gulp 3. Can be cleaned when Gulp 3 support is no longer required.

Closes #384

### DIFF
--- a/runner/gulp.ts
+++ b/runner/gulp.ts
@@ -20,24 +20,30 @@ import {test} from './test';
 
 
 export function init(gulp: Gulp, dependencies?: string[]): void {
+  // Compatibility with Gulp 3 and 4.
+  function compat(name, ...tasks) {
+    return gulp.series ? [name, gulp.series(...tasks)] : arguments;
+  }
+
   if (!dependencies) {
     dependencies = [];
   }
 
   // TODO(nevir): Migrate fully to wct:local/etc.
-  gulp.task('test', ['wct:local']);
-  gulp.task('test:local', ['wct:local']);
-  gulp.task('test:remote', ['wct:sauce']);
+  gulp.task.apply(gulp, compat('wct:local', dependencies, testLocal));
+  gulp.task.apply(gulp, compat('wct:sauce', dependencies, testRemote));
+  gulp.task.apply(gulp, compat('wct', ['wct:local']));
+  gulp.task.apply(gulp, compat('test:local', ['wct:local']));
+  gulp.task.apply(gulp, compat('test:remote', ['wct:sauce']));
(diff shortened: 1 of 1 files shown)


joaovieira:
CLA signed.

joaovieira:
Hm.. I thought Typescript supported spread operators. It didn't complain locally, but I'm also running on Node 6. Will have another look this weekend.

rictic:
Typescript does, but we're having it output ES6 until TS 2.1 comes with support for transporting async await to ES5, and so all es6 that's used needs to be supported by node v4

joaovieira:
To be honest, I actually find it better to define the task ourselves:

```javascript
import gulp from 'gulp'
import {test as wct} from 'web-component-tester'

gulp.task('test', () => {
  return wct({plugins: {local: {}, sauce: false}})
})
```

@rictic I will still make this PR Node 4 friendly, so you can then decide what to do :)

FredKSchott:
@joaovieira what is gained by doing this? Do our gulp tasks currently fail when run in gulp 4? Just curious as a gulp 4 noob

joaovieira:
@FredKSchott yes, the syntax is different. The dependency array in the second arg doesn't exist anymore. You need to build the pipeline yourself - using `gulp.series` and/or `gulp.parallel` (e.g. https://fettblog.eu/gulp-4-parallel-and-series).

This was more of a temporary fix until gulp 4 becomes the official release. Though, I'm now using the previous solution to define the task myself - https://github.com/Polymer/web-component-tester/pull/391#issuecomment-249354947

FredKSchott:
Got it, thanks @joaovieira. We're going to hold off on this and discuss gulp 4 support as a bigger change across all of our Polymer tooling repos. We'd like to keep all projects on the same version so that tooling devs don't run into trouble switching between repos. That might mean waiting until gulp 4 is officially released.

https://github.com/Polymer/web-component-tester/pull/391#issuecomment-249354947 should work for anyone who wants to use gulp 4 until then

Issue #383, opened 2016-09-07
prepare:webserver hook does not work how it is expected


sblancov:
I am developing a plugin for WCT and I am using something like this:

```
wct.hook('prepare:webserver', function (app, done) {
    app.use(bodyParse.json());
    app.use('/my-url', myFunction);
});
```

Running WCT v4.3.1 it works well, but WCT v4.3.4 fails. It throws error "app is not found". It seems the first parameter is "done" and "app" was disappeared :S

I think this hook should receive app in any version of WCT... but I have not found any information about it.

hypnoce:
The issue here is the conflict between the promise and callback form of emitHook.
In [context.ts](https://github.com/Polymer/web-component-tester/blob/d0128aa403232bc98dcf06a6b7071198eb8ef781/runner/context.ts#L116) the arguments length is expected to be greater then 2 because it always expects to have at least 2 parameters with the done callback to be the last parameter, which is not always true in the promise form.

In this particular case, it considers the Expression application to be the done callback. The promise form is used to [emit the 'prepare:webserver'](https://github.com/Polymer/web-component-tester/blob/d0128aa403232bc98dcf06a6b7071198eb8ef781/runner/webserver.ts#L143)

I'm not sure how to fix this since there are many solutions. Since all the hooks in wct are triggered using the Promise form, I would get rid of the callback form. Maybe @rictic has a better opinion on that.

gautaz:
I have the exact same issue when using the promise form:

wct.conf.js:
```javascript
module.exports = {
  registerHooks: (wct) => wct.hook('prepare:webserver', (app) => {
    console.log(app, app.use)
    return Promise.resolve()
  })
}
```

Executing the tests shows that `app` is a `[Function]` and obviously this is not the express application as `app.use` is `undefined`. As @hypnoce said, It seems `app` is passed the `done` function whatever the arguments length is and that there is no way to access the express application.

Is any workaround other than reverting back to WCT v4.3.1 available?

AdamKorczynski:
This is a duplicate bug, for workaround see https://github.com/Polymer/web-component-tester/issues/373

PR #399, opened 2016-10-09
Add empty callback to prepare:webserver

### LINKED ISSUE #373: express.use is not a function
using version 4.3.3

00:03:11.325 [14:32:56] Starting 'wct:local'...
00:03:12.239 Received {"global":{"statements":100,"branches":100,"lines":100,"functions":100}}
00:03:13.468 Installing and starting Selenium server for local browsers
00:03:18.775 Selenium server running on port 44707
00:03:18.801 [14:33:04] 'wct:local' errored after 7.47 s
00:03:18.801 [14:33:04] Error: express.use is not a function

Polymer 1.6
wct 4.3.3

AdamKorczynski:
This fixes problem where express middleware is not passed to hooks subscribed to perpare:webserver.

Fixes #373 and #383 

I would prefer to remove the callback option all together but that would break the wct-sauce dependency, which uses callbacks

### DIFF
--- a/test/unit/context.ts
+++ b/test/unit/context.ts
@@ -59,6 +59,19 @@ describe('Context', () => {
       expect(plugins).to.have.members(['local']);
     });
 
+    describe('hook handlers with only one argument', () => {
+      it('are passed the "done" callback function instead of the argument passed to emitHook', async() => {
+        const context = new Context();
+          context.hook('foo', function(arg1: any) {
+            expect(arg1).to.not.eq('hookArg');
+            // arg1 has been replaced with 'done'
+            arg1();
+        });
+
+        await context.emitHook('foo', 'hookArg');
+      });
+    });
+
     describe('hook handlers written to call callbacks', () => {
       it('passes additional arguments through', async() => {
         const context = new Context();
--- a/runner/webserver.ts
+++ b/runner/webserver.ts
@@ -140,7 +140,7 @@ export function webserver(wct: Context): void {
 
     // At this point, we allow other plugins to hook and configure the
     // webserver as they please.
-    await wct.emitHook('prepare:webserver', app);
+    await wct.emitHook('prepare:webserver', app, () => { });
 
     // Serve up all the static assets.
     app.use(serveWaterfall(wsOptions.pathMappings, {


kienpham2000:
@AdamKorczynski looks like you need to sign the cla before they can review this PR

AdamKorczynski:
Hmmm... not sure what I did wrong, will try to rectify. From what I can see it should be OK... From https://cla.developers.google.com/clas :

Contributor License Agreements
Individual Agreements

Individual agreements we have on file for you:

Agreement	Name	Date Signed	Manage
Google Individual CLA	Adam Korczynski	Oct 04, 2016 06:15 PDT	Edit Contact Information

AdamKorczynski:
Ooops, had wrong email address associated with the commit.
Amended and pushed.

lukasoppermann:
Please merge this, this fixes https://github.com/thedeeno/web-component-tester-istanbul/issues/37 so that I can get test coverage going. I tried it by manually applying the fix and I get the coverage report now.

test/unit/context.ts
```diff
@@ -59,6 +59,19 @@ describe('Context', () => {
       expect(plugins).to.have.members(['local']);
     });
 
+    describe('hook handlers with only one argument', () => {
+      it('are passed the "done" callback function instead of the argument passed to emitHook', async() => {
```
rictic: Is this a test that's documenting incorrect, buggy behavior? If so, please add a comment to that effect, or better yet, add it to the test name.

Did you take a look at how hard it would be to fix it instead?

rictic:
Thanks for the PR, and sorry for the slow response, the summit kinda ate 100% of our attention the past couple of weeks.

Yeah, the plan is to remove the callback option in an upcoming breaking change, but enough was changing in WCT with dropping node <4 and converting the project to typescript that I didn't want to add more instability on top.

justinfagnani:
I think the real bug is in the overload detection here: https://github.com/Polymer/web-component-tester/blob/2cce707c7e3e4a29ff06090f74297e26ec18eb4a/runner/context.ts#L116

when just called with a hook name and an app, that conditional doesn't trigger, even though it should. I think we need better logic to differentiate the (name, app) vs (name, done) cases.

justinfagnani:
I fixed `emitHooks` handling of arguments in https://github.com/Polymer/web-component-tester/pull/445

random82:
@justinfagnani @rictic are you able to cut an npm package with that change?

rictic:
I believe that it has landed in `@next` and we're actively working on landing it in v6: https://github.com/Polymer/web-component-tester/milestone/2

Issue #404, opened 2016-10-13
Paths .expand doesn't de-duplicate on Windows


AdamKorczynski:
paths .expand doesn't de-duplicate on Windows

When running the WCT unit tests on a non-POSIX OS several unit tests fail.

This is because the glob package doesn't accept forward-slash delimiters (https://www.npmjs.com/package/glob#windows) but node path will use the underlying OS' delimiter. Both will resolve */bar.js and *\bar.js to the same folder on POSIX and non-POSIX OS's, respectively, but a string comparison will fail.

It's a problem for unit test assertions, which are based on string comparison, and breaks the de-duplicating of files matched by a glob. De-duplicating fails because glob output is passed to a string comparison, _.union, in paths.expandDirectories.

I'll submit a PR.

FredKSchott:
Fixed via #405

PR #405, opened 2016-10-13
fix problem with globs not de-duplicating file matches on Windows OS …

AdamKorczynski:
paths .expand doesn't de-duplicate on Windows

This is because the glob package doesn't accept forward-slash delimiters but node path will use the underlying OS' delimiter. 

(#404)


FredKSchott:
Thanks for the PR @AdamKorczynski, and the thorough issue writeup!

FredKSchott:
I'm surprised we haven't gotten WCT set up on appveyor for windows testing. We should do that asap

PR #407, opened 2016-10-13
Updating mocha

maxknee:


FredKSchott:
@maxknee thanks for the contribution! We need to upgrade gulp-spawn-mocha as well to use mocha 3 & get rid of all install errors. I've created #408 to update both, lets work off of that branch going forward

PR #432, opened 2016-11-17
replace serve-waterfall with polyserve

FredKSchott:
***NOTE: Tests currently failing until we can get the latest polyserve version published.***

## Background

`wct._httpServer` is currently created as an express application during the "prepare" event, in a listener defined in `webserver.ts`. It serves files by duplicating a lot of the logic already found in polyserve. As its grown, the list of files and directories that it serves has become harder to define and reason about. 

Replacing this logic with polyserve will mean less code in wct, more confidence in the server logic, and (hopefully) less bugs.

## New Behavior

1. We are now using polyserve to create the express app and node server that will host our dependencies. As polyserve is expanded to handle multiple dependency versions, combinations and variants, wct will now also get these improvements.

## Removed Old Behavior

1. Serving a project's sibling directories as components. ***Currently in use in tedium.***
1. Serving WCT's dependencies as components. ***Currently in use in wct's own tests, and some 3rd-party projects*** You are now required to install any dependencies you need yourself, or use the `web-component-tester` bower package.
2. Serving sinon & lodash fallbacks via wct. You are now required to provide your own version of sinon and lodash if you use them in your app or your tests. This is a good thing.
3. Debug logging of where each requested was served from in the waterfall. This was more useful in the removed serve-waterfall, but it would still be a useful feature to add to polyserve.

## TODO

1. [ ] Merge this.
2. [ ] Add the ability to serve project siblings to polyserve, so that tedium will work on the latest wct.
3. [ ] Add request logging to polyserve.
4. [x] Look into next steps: Instead of returning a router, polyserve should become the entire server. Will need to be extensible so that wct can still add it's special features & fallbacks on top of it. Will need to coordinate with @rictic to make sure this aligns with his work.

---

/cc @justinfagnani @rictic


FredKSchott:
@rictic can you confirm that this usage of polyserve's `makeApp()` doesn't conflict with the work you're doing on the library?

justinfagnani:
Can you rebase without the formatting?

runner/webserver.ts
```diff
@@ -142,12 +143,13 @@ export function webserver(wct: Context): void {
...
 
-    // Serve up all the static assets.
-    app.use(serveWaterfall(wsOptions.pathMappings, {
+    // Serve up project & dependencies via polyserve
+    const polyserve = makeApp({
       root: options.root,
       headers: DEFAULT_HEADERS,
-      log: wct.emit.bind(wct, 'log:debug'),
-    }));
+      packageName: path.basename(options.root),
```
justinfagnani: packageName is normally read from `bower.json`. This matters in cases where there's a cycle in the dependency graph that points back to the project under test.

justinfagnani:
Why does this require a new version of polyserve? Are the tests only failing because of that? I would expect a bunch of tests to break, given wct had enough test coverage...

runner/webserver.ts
rictic: polyserve will also fall back to this if bower.json isn't there

FredKSchott:
@justinfagnani see broken tests, it's breaking on build, where polyserve breaks typescript compilation because it's custom_typings aren't referenced correctly. Fixed in latest version of polyserve.

runner/webserver.ts
FredKSchott: That's good to know about the default behavior. I'd rather keep this matching current wct behavior as much as possible and not add a dependency to bower for now, i'll remove the option and let polyserve use it's default.

runner/webserver.ts
```diff
@@ -142,12 +143,13 @@ export function webserver(wct: Context): void {
...
+    });
+    app.use('/components/', polyserve);
```
rictic: Won't this serve out the components dir at `/components/components/`?

FredKSchott: Check out how polyserve uses `makeApp()`: https://github.com/PolymerLabs/polyserve/blob/master/src/start_server.ts#L146 `makeApp()` returns a router that handles all routes on the `components/` path. This lets us continue to control the rest of the server that wct needs to manage. 

I've added as a todo to investigate what it would look like for polyserve to return a full express application, and for wct to build off of that. If possible, this would be the first step towards that.

FredKSchott: Good point, Updated my original comment to be more clear about what's happening

rictic:
Variants serving PR is at https://github.com/PolymerLabs/polyserve/pull/117

It doesn't modify makeApp because it needs to actually spawn web servers serving on distinct ports, so that API couldn't work. I can extend that PR to add a method that returns multiple apps for all of the different variants if that's what's needed here, but I wanted to get your feedback.

justinfagnani:
@rictic let's sync next week on the API for polyserve and how it integrates with wct and cli

FredKSchott:
This PR has been updated to use the variants branch of polyserve that @rictic has been working on. It's using a small fork of that branch that can be found here: https://github.com/PolymerLabs/polyserve/pull/120

WCT is working in every project I've run it on, but because we are no longer hosting wct's own dependencies as components the integration tests are still failing (see "removed old behavior" above). We'll need to do something to bring those dependencies used in the test fixtures into the fixtures themselves.

PR #424, opened 2016-11-03
Add the ability to test variant dependencies.

rictic:
- [x] CHANGELOG.md has been updated
 - Replaces #374

PR #436, opened 2016-12-01
Polyserve + Variants

rictic:

Supersedes #432 and #424

Depends on https://github.com/PolymerLabs/polyserve/pull/120

Integration tests are known failing.


CHANGELOG.md
```diff
@@ -1,14 +1,26 @@
-# 5.x
+## Unreleased
+
+### Added
+
+* Added first pass of _variants_. Variants different configurations of testing the same code.
+  * Add support for _variant dependencies_.
+    * wct already supports loading dependencies from your `bower_components` directory, mapping them to `../` in your code. You can now add variant dependency directories named like `bower_components-foo`. When these are detected, tests will then run separately for each such dependency directory, mapping `../` appropriately. See README for more details.
+
+### Removed
```
justinfagnani: Does this also remove wct auto-providing client-side dependencies via the npm package, ie: requiring the installation of the Bower package?

README.md
```diff
@@ -46,14 +46,10 @@ Your test suites can be `.html` documents. For example,
 </html>
 ```
 
-Note that it is _critical_ that you include `web-component-tester/browser.js` in
-your test suites. `browser.js` contains all of WCT's client logic (and loads
+Note that it is _critical_ that you include `../web-component-tester/browser.js`
```
justinfagnani: Does anyone use WCT on app-style projects where they'd want `/bower_components/web-component-tester/browser.js`

runner/browserrunner.ts
```diff
@@ -151,12 +159,10 @@ export class BrowserRunner {
   }
 
   startTest() {
-    const webserver = this.options.webserver;
-    const host = `http://${webserver.hostname}:${webserver.port}`;
-    const path = this.options.webserver.webRunnerPath;
-    const extra = (path.indexOf('?') === -1 ? '?' : '&') +
-        `cli_browser_id=${this.def.id}`;
-    this.browser.get(host + path + extra, (error) => {
+    const url = this.url;
```
justinfagnani: nit: not buying you much, since you only use it in one place. Either remove, or change the other `this.url`s.

runner/config.ts
```diff
@@ -52,19 +50,14 @@ export interface Config {
...
-    pathMappings: {[urlPath: string]: string}[];
-    // The URL prefix that serves contents from the project root.
-    urlPrefix: string;
-    webRunnerPath?: string;
-    webRunnerContent?: string;
+
     staticContent?: {[file: string]: string};
+
+    _generatedIndexContent?: string;
+    _servers?: {variant: string, url: string}[];
```
justinfagnani: Why is this optional? Don't we always start at least one server?

runner/steps.ts
```diff
@@ -135,22 +137,30 @@ function runBrowsers(context: Context) {
...
-          context.emit('log:debug', browser, 'BrowserRunner complete');
-        },
-        (error) => {
-          context.emit('log:debug', browser, 'BrowserRunner complete');
-          errors.push(error);
-        }));
-    return runner;
-  });
+
+  let runners: BrowserRunner[] = [];
```
justinfagnani: `const`? I don't see an assignment.

justinfagnani: btw, tslint landed a PR to add a `prefer-const` rule 2 days ago :)

runner/webserver.ts
```diff
@@ -32,21 +31,12 @@ const INDEX_TEMPLATE = _.template(fs.readFileSync(
     path.resolve(__dirname, '../data/index.html'), {encoding: 'utf-8'}));
 
 // We prefer serving local assets over bower assets.
```
justinfagnani: does this comment still make sense?

runner/webserver.ts
```diff
@@ -32,21 +31,12 @@ const INDEX_TEMPLATE = _.template(fs.readFileSync(
...
-const SERVE_STATIC = {
-  // Keys are regexps.
-  '^(.*/web-component-tester|)/browser\\.js$':
-      path.join(PACKAGE_ROOT, 'browser.js'),
-  '^(.*/web-component-tester|)/browser\\.js\\.map$':
-      path.join(PACKAGE_ROOT, 'browser.js.map'),
-  '^(.*/web-component-tester|)/data/a11ySuite\\.js$':
-      path.join(PACKAGE_ROOT, 'data', 'a11ySuite.js'),
-};
+const WCT_ROOT = path.resolve(__dirname, '..');
```
justinfagnani: is this used?

runner/webserver.ts
```diff
@@ -82,112 +72,136 @@ export function webserver(wct: Context): void {
...
+          options.root, 'bower_components', 'web-component-tester',
+          'browser.js');
+      send(request, browserJsPath).pipe(response);
     });
+    // TODO(rictic): detect that the user hasn't bower installed wct and die.
 
-    // Mapped static content (overriding files served at the root).
     _.each(wsOptions.staticContent, function(file, url) {
-      app.get(new RegExp(url), function(request, response) {
+      additionalRoutes.set(url, (request, response) => {
```
justinfagnani: what's in staticContent? Do they deserve warnings?

runner/webserver.ts
```diff
@@ -82,112 +72,136 @@ export function webserver(wct: Context): void {
...
-        response.set(DEFAULT_HEADERS);
-        response.send(wsOptions.webRunnerContent);
-      });
-    }
-
-    // At this point, we allow other plugins to hook and configure the
-    // webserver as they please.
-    await wct.emitHook('prepare:webserver', app);
+    const pathToGeneratedIndex =
+        `/components/${path.basename(options.root)}/generated-index.html`;
```
justinfagnani: path.basename(options.root) is not guaranteed to be the package name that polyserve uses. Polyserve tries to use the bower package name first. Does this file need to appear to come from the package under test? Can we just serve it from a special top-level path?

runner/webserver.ts
```diff
@@ -82,112 +72,136 @@ export function webserver(wct: Context): void {
...
 
-    // Serve up all the static assets.
-    app.use(serveWaterfall(wsOptions.pathMappings, {
+    // Serve up project & dependencies via polyserve
+    const polyserveResult = await startServers({
       root: options.root,
       headers: DEFAULT_HEADERS,
-      log: wct.emit.bind(wct, 'log:debug'),
-    }));
+      packageName: path.basename(options.root),
```
justinfagnani: why wouldn't we use the package name from bower.json?

runner/webserver.ts
```diff
@@ -82,112 +72,136 @@ export function webserver(wct: Context): void {
...
+      additionalRoutes: additionalRoutes,
+    });
+    let servers: Array<MainlineServer|VariantServer>;
+    if (polyserveResult.kind === 'mainline') {
```
justinfagnani: I'd still much rather have polyserveResult be a consistent type and not have to hoist this kind of conditional on clients:

This should work no matter how many servers polyserve starts:

```typescript
servers = polyserveResult.servers;
wsOptions.port = polyserveResult.mainline.server.address().port;
```

runner/webserver.ts
rictic: We explicitly tell polyserve to use this packageName, so it won't infer it from the bower.json.

Refactoring this to serve somewhere else would be possible, but would involve touching a bunch of stuff and wouldn't actually solve the problem, as we'd need to know the path to the suites to load, which will be served out from the project under test, so we'd need to compute that path one way or another.

Clarified in the code a bit what's going on.

runner/webserver.ts
rictic: We could, but then we'll need to write the code to try to load it from bower.json, gracefully fall back, etc. How valuable is this?

runner/webserver.ts
rictic: Yeah, I think you're right. Fix when we refactor to always launch a control server? The change will be backcompat, as we'll just never be returning an object with a kind of 'mainline'

runner/webserver.ts
justinfagnani: Not sure how valuable in practice, but the reason why polyserve defaults to the bower package name is that it's possible for there to be dependency cycles back onto the package under test, in which case we want another packages `../this-package/foo` import to work.

runner/webserver.ts
justinfagnani: sgtm

justinfagnani:
If everything's addressed, then this is probably good to go now, but I'm going to do one more pass in a bit.

runner/webserver.ts
```diff
@@ -12,54 +12,37 @@
...
 // https://docs.saucelabs.com/reference/sauce-connect/#can-i-access-applications-on-localhost-
 // - 80, 443, 888: these ports must have root access
 // - 5555, 8080: not forwarded on Android
-const SAUCE_PORTS = [
-  2000, 2001, 2020, 2109, 2222, 2310, 3000, 3001, 3030, 3210, 3333,  4000,
-  4001, 4040, 4321, 4502, 4503, 4567, 5000, 5001, 5050, 5432, 6000,  6001,
-  6060, 6666, 6543, 7000, 7070, 7774, 7777, 8000, 8001, 8003, 8031,  8081,
-  8765, 8777, 8888, 9000, 9001, 9080, 9090, 9876, 9877, 9999, 49221, 55001
-];
+// const SAUCE_PORTS = [
```
justinfagnani: what's up with the comments?

runner/webserver.ts
```diff
@@ -75,119 +58,129 @@ export function webserver(wct: Context): void {
...
+    // webservers as they please.
+    for (const server of servers) {
+      await wct.emitHook('prepare:webserver', server.app);
+    }
 
-    app.get('/favicon.ico', function(request, response) {
-      response.end();
+    options.webserver._servers = servers.map(s => {
+      return {
+        url: `http://localhost:${s.server.address().port}${pathToGeneratedIndex
```
justinfagnani: ugh, clang :(

runner/webserver.ts
```diff
@@ -75,119 +58,129 @@ export function webserver(wct: Context): void {
...
+      };
     });
 
-    app.use(function(request, response, next) {
-      wct.emit('log:warn', '404', chalk.magenta(request.method), request.url);
-      next();
-    });
+    // TODO(rictic): re-enable this stuff. need to either move this code into
+    //     polyserve or let the polyserve API expose this stuff.
+    // app.use('/httpbin', httpbin.httpbin);
```
justinfagnani: is this not handed to additionalRoutes?

runner/webserver.ts
rictic: Moved this info to https://github.com/PolymerLabs/polyserve/issues/134

runner/webserver.ts
rictic: Cleaned up.

package.json
```diff
@@ -79,7 +79,7 @@
     "mocha": "^3.1.2",
     "multer": "^1.1.0",
     "nomnom": "^1.8.1",
-    "polyserve": "^0.16.0-prerelease.3",
+    "polyserve": "0.16.0-prerelease.4",
```
justinfagnani: I just published `.5`

rictic: It didn't include the updated type defs

runner/webserver.ts
rictic: It isn't. httpbin is a Router, not a normal request handler, so it does its own sub-path matching. I tried a few things to get this to work with additionalRoutes and nothing seemed to work.

Filed as https://github.com/PolymerLabs/polyserve/issues/138

package.json
rictic: Using .6

rictic:
merged!

PR #442, opened 2016-12-07
Fail if wct isn't installed in project under test

### LINKED ISSUE #439: Detect that wct hasn't been installed in the package
… and die with a helpful error message

rictic:
Fixes #439


runner/webserver.ts
```diff
@@ -63,6 +63,15 @@ export function webserver(wct: Context): void {
     const wsOptions = options.webserver;
     const additionalRoutes = new Map<string, RequestHandler>();
 
+    const pathToLocalWct =
+        path.join(options.root, 'bower_components', 'web-component-tester');
+    if (!exists(pathToLocalWct)) {
+      throw new Error(`WCT isn't installed locally. Run:
+    bower install --save-dev web-component-tester
```
justinfagnani: I think we should expand this message. Maybe something like:

```
The web-component-tester Bower package is not installed as a dependency of this project.

Please run this command to install:
    bower install --save-dev web-component-tester

Web Component Tester 6.0 now requires that support files needed by the browser are installed as part of the project's dependencies or dev-dependencies. This is so that projects are ensured to get the correct version of support files to match their version of web-component-tester.
```

justinfagnani:
Do we need to check for the npm package as well? Once a project installs the Bower package, they are relatively fixed to that version, but if they don't have the npm package they're at the whim of the globally installed version. I'm not aware of any incompatibilities, but after doing this we could see unusual combos of newer and older npm packages with 6.x bower packages.

rictic:
If we check for the npm package we'll also need to check that their versions are compatible yeah?

runner/webserver.ts
rictic: +1. Tweaked it a bit, PTAL

justinfagnani:
Yeah. Ideally we'd have a semver range stashed in the npm package's package.json that we could match against the Bower version. Failing that we could check for major.minor compatibility and assume we don't rely on patch behavior.

rictic:
I'm unsure about whether we should require wct to be installed via npm locally as well. Moved discussion of that to https://github.com/Polymer/web-component-tester/issues/444

rictic:
Stashed a semver of the allowed client-side versions and checked against it. PTAL!

PR #443, opened 2016-12-07
Fix compilation in typescript 2.1

rictic:


runner/config.ts
```diff
@@ -436,17 +436,20 @@ export async function expand(context: Context): Promise<void> {
 function expandDeprecated(context: Context) {
   const options = context.options;
   // We collect configuration fragments to be merged into the options object.
-  const fragments: {plugins: {sauce: {}, local?: {}}}[] = [];
+  const fragments = [];
```
rictic: This can now be properly inferred.

runner/config.ts
```diff
@@ -436,17 +436,20 @@ export async function expand(context: Context): Promise<void> {
...
   browsers = <any>_.compact(<any>browsers);
   if (browsers.length > 0) {
     context.emit(
-        'log:warn', 'The --browsers flag/option is deprecated. Please use ' +
+        'log:warn',
+        'The --browsers flag/option is deprecated. Please use ' +
             '--local and --sauce instead, or configure via plugins.' +
             '[local|sauce].browsers.');
-    const fragment = {plugins: {sauce: {}, local: {}}};
+    const fragment: {plugins: {[name: string]: {browsers?: Browser[]}}} = {
```
rictic: This now needs to be told that plugins can have arrays of browsers.

justinfagnani:
Shouldn't this include a bump of the typescript version to ^2.1.0?

rictic:
Good point, done

runner/config.ts
justinfagnani: That's a little much inference, IMO, but hey. I hope they're keeping a spec.

PR #450, opened 2016-12-13
.replace no longer relies on Polymer internals

e111077:

The `.replace` function no longer relies on Polymer internals. Instead of stubbing `Polymer.Base.instanceTemplate` or `Polymer.Element.prototype._stampElement`, it just stubs `document.importNode` (with the blessing of Kevin). Tested on both Polymer 1.x and 2.x.

NOTE: The tests for it do rely on Polymer internals because they go through usual Polymer use cases, but the tests are compatible with both Polymer 2.x and 1.x. LMK if this is a problem; can change to webcomponent lifetime tests if preferred.


justinfagnani:
Looks good. Are there any docs that need to be updated? Either in this repo or on the site?

e111077:
API is the same. I have noted the stubbing of `document.importNode` in the changelog; I can't think of anything else to document.

PR #451, opened 2016-12-13
Extract test fixture

e111077:
- [x] changelog updated

Removed `test-fixture` from WCT. See [here](https://github.com/PolymerElements/test-fixture#even-simpler-usage-in-mocha) for mocha integration.


fearphage:
Was there a reason or benefit to this? I'm just curious. Thanks.

justinfagnani:
Including test-fixture in WCT makes it difficult to have tests that use different versions of test-fixture itself

fearphage:
Excellent. Thanks!

PR #452, opened 2016-12-13
Deep merge of browser options

Ado77:
Made the browser options do a deep merge instead of a shallow one to allow more options through


Ado77:
I signed it

PR #454, opened 2016-12-13
Deep merge of browser options

Ado77:
Made the browser options do a deep clone instead of a shallow one to allow more options through


random82:
That will allow overriding default wct-local options from the plugin configuration

Ado77:
I know everyone is focused on the new version of the components but this would unblock our CI pipeline... Is there any chance  :smile_cat: @notwaldorf :smile_cat:, @justinfagnani or @rictic could have a quick look?

/End of long winded bump

justinfagnani:
I'll take a look

notwaldorf:
Hah, thanks for the extra cat emoji! (I'm useless at wct so I can't help though)

PR #457, opened 2016-12-15
Warn if wct is not installed locally in node_modules.

### LINKED ISSUE #444: Warn if there is no local npm install of wct in project under test
In wct 6.0 we're moving to requiring that the project under test has a local install of the client-side code for wct. In the current bower world that means that they need to have run `bower install web-component-tester`. This reduces magic and gives the project under test more control over the versions of their test-time dependencies.

Should we do the same for the node side of wct, and require that the project under test have wct installed in their node_modules directory? Already the global `wct` command defers to any local installation, if any. This change would reduce the impact of any future breaking changes in wct, but it would increase the friction of getting started with wct, and require users to use both bower and npm in all of their projects under test.

rictic:

Fixes #444

### DIFF
--- a/runner/webserver.ts
+++ b/runner/webserver.ts
@@ -118,6 +118,9 @@ Expected to find a ${mdFilenames.join(' or ')} at: ${pathToLocalWct}/
     We recommend that you run this command to add it:
         npm install --save-dev web-component-tester
 
+    or run:
+        yarn add web-component-tester --dev
+
     Doing so will ensure that your project is in control of the version of wct
     that your project is tested with, insulating you from any future breaking
     changes and making your test runs more reproducible. In a future release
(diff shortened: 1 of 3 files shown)


runner/webserver.ts
```diff
@@ -101,7 +104,27 @@ Expected to find a ${mdFilenames.join(' or ')} at: ${pathToLocalWct}/
...
+    // Check that there's a wct node module.
+    const pathToWctNodeModule =
+        path.join(options.root, 'node_modules', 'web-component-tester');
+    if (!exists(pathToWctNodeModule)) {
+      throw new Error(`
+    The web-component-tester node module is not installed as a dependency of
+    this project (${packageName}).
+
+    Please run this command to install:
+        npm install --save-dev web-component-tester
```
justinfagnani: Can you add `yarn add web-component-tester --dev`?

rictic: Done

PR #462, opened 2016-12-16
Update client-side-version-range with new prerelease tag

rictic:


rictic:
Merging this in TBR as it's a tiny change and travis is broken without

PR #463, opened 2016-12-16
Use polyserve's ESNext -> ES5 compilation.

### LINKED ISSUE #437: Add --compile flag
Pass through to polyserve, for auto/forced/never compilation to ES5.

rictic:

Also hook the `webserver.hostname` flag back up, as it turns out it's used.

Fixes #437

/cc @azakus

### DIFF
--- a/runner/config.ts
+++ b/runner/config.ts
@@ -54,6 +54,9 @@ export interface Config {
     // determined at runtime if none is provided.
     port: number;
 
+    // The hostname used when generating URLs for the webdriver client.
+    hostname: string;
+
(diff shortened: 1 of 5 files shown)


runner/config.ts
```diff
@@ -217,6 +225,13 @@ const ARG_CONFIG = {
   // Managed by supports-color; let's not freak out if we see it.
   color: {flag: true},
 
+  compile: {
+    help: 'Whether to compile ES>=6 down to ES5. ' +
```
justinfagnani: It's not ES>=6, it's exactly ES2015

PR #467, opened 2016-12-21
Update the client version range

rictic:
This works, our integration tests cover this.


rictic:
Huh, something's wrong with compiling at head, checking on that.

PR #489, opened 2017-02-08
update findup-sync & glob to remove deprecation warnings

FredKSchott:
- findup-sync: `v0.2` -> `v0.4` (no breaking changes)
- glob: `v5` -> `v7` (small breaking change, described below)

This gets rid of the DDoS deprecation warning on install of Polymer CLI & WCT by updating two outdated uses of glob `v5`.

The only change to behavior is that `!xxx` glob patterns in `loadSuites()` now won't do anything, instead of incorrectly matching all files. Since this behavior was already broken, this shouldn't constitute a breaking change for WCT.


rictic:
Nice, glad to see these warnings going away

PR #490, opened 2017-02-10
Upgrade to lodash 4.

### LINKED ISSUE #342: upgrade lodash from '3.10.1'
hi, this is an old version, pls do upgrade.

sankethkatta:
Fixes #342 

 - [X] CHANGELOG.md has been updated


rictic:
Lodash can be a painful dependency to for users to update. It would be better and much less disruptive if we instead made wct compatible with either lodash v3 or v4. That way users could decide for themselves, by specifying their preferred version in their bower.json.

sankethkatta:
@rictic That sounds fair. We could either leave the code written against v3 and create a small shim for v4 to have backwards compatibility. Or we could do the opposite and migrate the code to v4 and write the shim to make v3 forwards compatible. 

Any thoughts on which direction would be better?

rictic:
If the two changes here are the only ones needed, what if we rewrote those expressions to not use lodash at all?

sankethkatta:
It seems that I broke some other things by upgrading the packages. The CI is exiting successfully, but there is no test result output shown. If it turns out to be just a few places, we can definitely just rewrite them to not use lodash.

slhenty:
Love to see this sooner than later... 👍 
I'm blocked on using sinon.fakeTimers until I can easily use lodash 4.x.

rictic:
We likely won't get to this in the next month, as we're all hands on deck for the Polymer 2.0 release. PRs that remove wct's use of lodash would be very much appreciated though.

sankethkatta:
Sorry for the delay, but I can't seem to find what the issue is with upgrading to lodash 4.x. The test output just stops displaying once I upgrade / rename missing methods. I can't tell if its the actual code breaking or the test runner. Sounds like the effort would be better spent on eliminating the lodash dependency all together though.

stramel:
Can we instead remove the lodash dependency?

justinfagnani:
@stramel I'd love to remove all client-side magically included libraries. Tests can bring their own libraries as needed, at whatever versions they need, and WCT would be able to evolve much easier. We're very constrained at the moment.

AlphaWong:
The conflicts Jas even resolved ?

hastebrot:
Here is a list of current versions:
- mocha: 4.0.0 / 2017-10-03
- chai: 4.1.2 / 2017-08-31
- sinon: 4.0.1 / 2017-10-04
- lodash: 4.17.4 / 2017-01-06

FredKSchott:
Auto-closing this PR as abandoned. This happens if there hasn't been any activity on a PR in a long time.

@sankethkatta if you are still interested in getting this PR merged: please rebase, and then ping me to reopen and I'll be happy to take a look.

Issue #392, opened 2016-09-28
a11ySuite should allow for imperative setup steps


azakus:
If an element has some lazily created DOM, there is currently no way to test this with `a11ySuite`.

azakus:
It might be enough to give a11ySuite a callback that the user can control.

usergenic:
Should be fixed via #493 `->` #513

PR #493, opened 2017-02-15
Adds Polymer.dom.flush() and beforeEach to a11ySuite

lukeschaefer:
As per issue #392, parts of the page may be missing while a11ySuite runs, since Polymer.dom.flush() is never called.

Without Polymer.dom.flush, fixtures with elements containing dom-repeat or dom-if templates would not be fully loaded. In many cases, this results in running the test suite on only a fragment of the page, missing possible errors. In few cases, it would trigger possible errors, for example:

    <div role='radiogroup'>
      <template is='dom-repeat'>
         <ul role='radio'>Button</ul>
      </template> 
    </div>

The test suite would incorrectly think the radiogroup had no role='radio' children.

Also, this adds in a beforeEach function that can be passed in, allowing any extra setup to run before each test.


lukeschaefer:
Comment to sync up the googlebot

usergenic:
Looks good.  I'm going to merge this into a repo-local branch to make some stylistic changes before merging to master.

Thanks!

PR #497, opened 2017-02-27
New week, new test-fixture version

notwaldorf:
It's a brave new world out there. 

@justinfagnani @cdata


notwaldorf:
Tests are green now, merging!

PR #512, opened 2017-03-21
listen only to WebComponentsReady

valdrinkoshi:
⚠️ potentially breaking change ⚠️ 
WCT should just wait for `WebComponentsReady` by default. If user requires different wait time, he should set the `WCT.waitFor` function. 

WCT currently does too many checks to see if HTMLImports is defined, or Polymer.ready is defined. In particular, if `HTMLImports` is not defined, WCT would synchronously execute the tests. This is a problem when we use `webcomponents-loader.js` as it will asynchronously load the polyfills.
```html
<script src="webcomponents/webcomponents-loader.js"></script>
<script src="web-component-tester/browser.js"></script>
```
In chrome, HTMLImports will never be defined, so tests will be executed too early.

Needed by https://github.com/webcomponents/webcomponentsjs/pull/742


browser/util.js
```diff
@@ -15,56 +15,23 @@ import * as config from './config.js';
...
  */
 export function whenFrameworksReady(callback) {
-  debug('whenFrameworksReady');
-  var done = function() {
-    debug('whenFrameworksReady done');
+  // webcomponents-loader.js will eventually fire WebComponentsReady.
+  // If WebComponents is not defined and we cannot find the script,
+  // invoke the callback.
+  if (!window.WebComponents &&
+    !document.querySelector('script[src*="webcomponents-loader.js"]')) {
```
rictic: What about other files like webcomponents-es5-loader.js, webcomponents-es6-loader.js (proposed renaming of webcomponents-loader.js), webcomponents-lite.js, or webcomponents.js?

valdrinkoshi: Re es5/es6 loaders: the idea is to have only `webcomponents-loader.js` (see more info [here](https://github.com/webcomponents/webcomponentsjs/issues/737#issuecomment-288191881)).

`webcomponents-lite.js` and `webcomponents.js` both set `window.WebComponents` and fire `WebComponentsReady` event.

Here we need a way to know if to wait for the WebComponentsReady event or not. Another approach is to always wait for the event for a maximum of 1 second (or any other arbitrary time) and then execute the tests, but that's not any better than this :/

valdrinkoshi:
This could potentially be a breaking change for users who rely on WCT to start tests right **after** `HTMLImports.whenReady(callback)` and **before** `WebComponentsReady` event gets fired.

I'm not familiar with the timings of `Polymer.whenReady(callback)` (Polymer 0.5 readiness signal) or `WebComponents.whenReady(callback)`, I'm assuming they're the same as when `WebComponentsReady` event is fired.

browser/util.js
rictic: When is whenFrameworksReady called? Can we wait for all scripts to finish loading first, so that we don't have to look for script tags with special srcs?

valdrinkoshi:
@rictic another way to fix this w/o introducing breaking changes is changing [line 56](https://github.com/Polymer/web-component-tester/blob/master/browser/util.js#L56) to:
```js
if (!window.HTMLImports) {
  requestAnimationFrame(done);
}
```
WDYT?

valdrinkoshi:
@rictic I realized that the issue I'm facing happens because of the new `webcomponents-loader.js` which will load the right bundle asynchronously. 
Normally people use `webcomponents-lite.js` in their tests. Since I'm using `webcomponents-loader.js`, it's my duty to say WCT to wait for it, e.g.
```js
    // Tell WCT to wait for WebComponentsReady.
    window.WCT = {
      waitFor: function(cb) {
        addEventListener('WebComponentsReady', cb);
      }
    };
```
I vote for closing this PR and ask people to do the same if they want to use `webcomponents-loader.js` in their tests. WDYT?

rictic:
What is the alternative to using webcomponents-loader in tests?

valdrinkoshi:
Using `webcomponents-lite.js`, but actually this is something we're trying to move away from, and encourage people to always use `webcomponents-loader.js`.

Currently we've opted for stubbing HTMLImports in [webcomponents-loader.js](https://github.com/webcomponents/webcomponentsjs/blob/master/webcomponents-loader.js#L36) so that WCT will wait for `WebComponentsReady`, so this PR is not needed.

PR #513, opened 2017-03-22
Add `Polymer.dom.flush()` before each test.

usergenic:
- This is a follow-up to https://github.com/Polymer/web-component-tester/pull/493
 - @lukeschaefer's original PR description:
> As per issue #392, parts of the page may be missing while a11ySuite runs, since `Polymer.dom.flush()` is never called.
>
>Without Polymer.dom.flush, fixtures with elements containing dom-repeat or dom-if templates would not be fully loaded. In many cases, this results in running the test suite on only a fragment of the page, missing possible errors. In few cases, it would trigger possible errors, for example:
>
> ```html
> <div role='radiogroup'>
>   <template is='dom-repeat'>
>      <ul role='radio'>Button</ul>
>   </template> 
> </div>
> ```
> The test suite would incorrectly think the radiogroup had no role='radio' children.
>
> Also, this adds in a beforeEach function that can be passed in, allowing any extra setup to run before each test.

### DIFF
--- a/data/a11ySuite.js
+++ b/data/a11ySuite.js
@@ -51,13 +51,15 @@
           a11ySuite.eachTest = function() {
             // instantiate fixture
             fixtureElement.create();
-            
+
             // Make sure lazy-loaded dom is ready (eg <template is='dom-repeat'>)
             Polymer.dom.flush();
-            
+
             // If we have a beforeEach function, call it
-            if(beforeEach) beforeEach();
-            
+            if (beforeEach) {
+              beforeEach();
+            }
+
             // run audit
             var auditResults = axs.Audit.run(axsConfig);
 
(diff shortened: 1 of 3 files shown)


data/a11ySuite.js
```diff
@@ -50,7 +51,13 @@
           a11ySuite.eachTest = function() {
             // instantiate fixture
             fixtureElement.create();
-
+            
+            // Make sure lazy-loaded dom is ready (eg <template is='dom-repeat'>)
+            Polymer.dom.flush();
+            
+            // If we have a beforeEach function, call it
+            if(beforeEach) beforeEach();
```
rictic: Add braces to one line ifs please. It's more robust to merges, typos, ets. One `goto fail;` per civilization is enough I think :)

usergenic:
Thanks for the review, @rictic!

PR #519, opened 2017-03-31
clone and replace non upgraded Custom Elements

### LINKED ISSUE #518: [WCT v5.0.1] test-fixture not upgraded when injected in Chrome 57
Looks like the fix shipped in https://github.com/Polymer/web-component-tester/pull/506 doesn't work when using `document.registerElement` (which is done by `test-fixture v1.x`). This is a Chrome bug https://bugs.chromium.org/p/chromium/issues/detail?id=701601#c21

No easy workaround I can think of so far :/

valdrinkoshi:
Fixes #518, #505.
PR #506 did provide a fix that worked only when Custom Elements are declared via customElements.define. This PR fixes also the case when they're defined via document.registerElement.

`CHANGELOG.md` doesn't need an update (is already mentioning the workaround).

This change needs to be shipped in 6.x and 5.x


browser/suites.js
```diff
@@ -114,7 +114,17 @@ export function runSuites(reporter, childSuites, done) {
 function _runMocha(reporter, done, waited) {
   if (config.get('waitForFrameworks') && !waited) {
     var waitFor = (config.get('waitFor') || util.whenFrameworksReady).bind(window);
-    waitFor(_runMocha.bind(null, reporter, done, true));
+    waitFor(function() {
+      // NOTE: In Chrome57 test-fixture elements in the document might not get upgraded when
+      // there is a high GC https://bugs.chromium.org/p/chromium/issues/detail?id=701601
+      // We clone and replace the ones that weren't upgraded.
+      var tfs = document.querySelectorAll('test-fixture');
```
rictic: Is test-fixture the element we should be concerned about? I'd say we should just crawl the entire document for unupgraded custom elements.

browser/suites.js
```diff
@@ -114,7 +114,17 @@ export function runSuites(reporter, childSuites, done) {
...
+      for (var i = 0, l = tfs.length, tf; i < l && (tf = tfs[i]); i++) {
```
rictic: Let's go with simpler and more legible for loops, even if it adds a few more expressions to the loop body. There's just so many subtle ways to mistype this sort of code and have it almost work

browser/suites.js
```diff
@@ -114,7 +114,17 @@ export function runSuites(reporter, childSuites, done) {
...
+        if (typeof tf.create === 'function') continue; // This one was upgraded!
+        tf.parentNode.replaceChild(document.importNode(tf, true), tf);
```
rictic: clever!

browser/suites.js
valdrinkoshi: Good point, will update this. I think I'll also detect if we're in Chrome 57 & bail out if not, WDYT?

rictic: +1

valdrinkoshi:
@rictic feedback incorporated, WDYT?

PR #522, opened 2017-04-03
limit creating new instances

valdrinkoshi:
### Update

It looks like the issue https://bugs.chromium.org/p/chromium/issues/detail?id=701601 cannot be reliably worked around. The GC might kick at any time, multiple times. Cloning and replacing non-upgraded instances before running tests doesn't handle the cases where the GC kicks later in time (e.g. when elements are detached).

----

Refinement of the workaround done in https://github.com/Polymer/web-component-tester/pull/519

The constructor of Custom Elements that are declared via `document.registerElement` cannot be easily retrieved. The safest check is to create a new instance and get its constructor.

If an element does things on `createdCallback`, creating new elements might lead in potential errors (e.g. singleton elements setting global variables on `window` when created).

So the strategy is to check if the instance has `createdCallback` already defined.
If an element doesn't do things on `createdCallback` but in some other lifecycle callback, it is "fine" to create a new instance.

Requires to be shipped in 6.x and 5.x


valdrinkoshi:
This needs further work. Cloning and reparenting has bad consequences ;_;

PR #533, opened 2017-04-19
Wait for WebComponentsReady

valdrinkoshi:
Change default `waitFor` to rely only on `WebComponentsReady` event to know if we should wait for the frameworks. We wait for the event if there is a script with src containing "webcomponents"

- [x] removed `webcomponents.min` script from `nested` tests which weren't relying on that anyways, and they had wrong paths

To be merged in 6.0.0 only
Needed in https://github.com/webcomponents/webcomponentsjs/pull/762


valdrinkoshi:
Finally green tests 💚 
@rictic @sorvell @kevinpschaaf PTAL

PR #508, opened 2017-03-15
Prepare to release v5.0.1

rictic:
Includes a cherry-pick of 7d0c9d6f

 - [x] CHANGELOG.md has been updated

PR #536, opened 2017-04-21
Remove _fixCustomElements workaround

valdrinkoshi:
Reverts workarounds we tried for the [GC issue in Chrome 57](https://bugs.chromium.org/p/chromium/issues/detail?id=701601).

Through those workarounds we were able to ensure custom elements were upgraded before starting the tests, but in complex test scenarios GC was kicking in again, and it's impossible for WCT to help on those. Hence, it's better to remove this code from WCT and expect the user of WCT to test on Chrome !== M57 (that's what is done by polymer elements already).

The workaround is present in v6.0.0 (#519 and #506) and v5.0.1 (#508). Reverting it from 5.x (if needed) should be done in a separate PR.

- [x] CHANGELOG.md updated


stramel:
Looks like this will need a rebase before merging

valdrinkoshi:
Done 👌

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
   * Upgraded @types/gulp
   * Moved all @types to devDependencies
   * Removed PolymerElements/test-fixture from npm dependencies (is already installed by bower)
+  * Undo fix for #505, as https://bugs.chromium.org/p/chromium/issues/detail?id=701601 has been fixed and shipped in M58 stable.
```
stramel: NIT: This should actually be unindented 2 spaces. The above bullets are a sublist of "Updated package.json dependencies".

valdrinkoshi: True! Will update

valdrinkoshi:
Build is still failing, and I have no clue why. Seems it complains about the gulpfile linting?
https://travis-ci.org/Polymer/web-component-tester/jobs/245905352
@justinfagnani @stramel any ideas?

justinfagnani:
Looks like we probably picked up some breaking type definition changes :/

valdrinkoshi:
That wasn't caused by this PR though, right? I'm not sure which dependency's latest version causes those errors. We should probably lock the versions to avoid this in the future?

stramel:
@valdrinkoshi Correct, it wasn't from your PR. The overall build is failing on this project due to those linting errors. I have hacked around them in my PR #568 but I'm not sure if we want to do that or try to properly fix them.

PR #590, opened 2017-08-16
Add a '--npm' flag to support client code based on npm (instead of bower)

usergenic:
- Added an `--npm` flag to support web packages which are installed in `node_modules` by npm (instead of installed in `bower_components` by Bower.
 - Please note this PR is merging to `npm-preview` branch which will not be merged into master yet because it is experimental preview.


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "web-component-tester",
-  "version": "6.0.1",
+  "version": "6.0.1-dev.1",
```
justinfagnani: `6.1.0-pre.1` maybe?

runner/webserver.ts
```diff
@@ -179,9 +186,11 @@ Expected to find a ${mdFilenames.join(' or ')} at: ${pathToLocalWct}/
     }
 
     options.webserver._servers = servers.map(s => {
+      const npmParam = options.npm ? `?npm=true` : '';
       const port = s.server.address().port;
+      const url = `http://localhost:${port}${pathToGeneratedIndex}`;
       return {
-        url: `http://localhost:${port}${pathToGeneratedIndex}`,
+        url: `${url}${npmParam}`,
```
justinfagnani: isn't there another query parameter added for the browser id? Does this cause double `?` in the url?

usergenic: In the initial URL generation there are no parameters and when `cli_browser` is added there's a check to add via ? or &

Issue #602, opened 2017-08-31
WCT hangs when running test on IE11 (Windows 10)


fcontigiani-namirial:
web-component-tester/browser/config.js, line 13:

var useNpm = document.currentScript.src.match(/wct-browser-legacy\/browser.js/) // ...

document.currentScript is undefined on IE11

gwicks:
Having the same issue here

usergenic:
PR #603 removes use of currentScript for this and should make things work again on IE11.  Working on getting this released.

usergenic:
web-component-tester 6.1.4 published 🎉

usergenic:
please update your wct

usergenic:
issue not completely fixed.  there's a follow-up release coming shortly.

usergenic:
wct 6.1.5 and polymer-cli 1.5.4 published 🎉 <- *actually* fixes this

PR #603, opened 2017-08-31
Remove reliance on document.currentScript

usergenic:
- Removing recently added dependency on `document.currentScript` which is unavailable in IE11
 - Fixes https://github.com/Polymer/web-component-tester/issues/602


rictic:
So, this is a change in behavior though, yeah?

rictic:
Instead of loading wct's browser.js with an `npm` query param, now the toplevel page needs to be loaded with the `npm` query param. Is this documented somewhere? I think it should definitely be in the changelog.

usergenic:
@rictic I added behavior that checked for "npm mode" by incorrectly choosing to use document.currentScript -- That check itself breaks in IE11 and people that could care less about npm mode are effected.

usergenic:
The top-level page is loaded with the npm query param by wct.  That behavior already happened.

usergenic:
I'll put the note in the changelog, but we havent really covered anything wrt npm testing with wct yet... That is forthcoming

rictic:
So previously you would both load the toplevel page with `?npm=true` and you would load browser.js as `browser.js?npm=true`?

rictic:
Ah ok, yeah a line in the changelog is enough then. I'm not sure I 100% get the change, but I'm not blocking this PR from landing.

PR #610, opened 2017-09-05
Updated browser.js to use test-fixture as js import in the npm case.

usergenic:
- `@polymer/test-fixture` was incorrectly still loaded as html import-- converted to script include.
 - Updated wct-browser-legacy package.json to version 0.0.1-pre.7 (and published).


usergenic:
Please note appveyor doesn't work for WCT yet so you can ignore those CI fails.

@FredKSchott PTAL

PR #616, opened 2017-09-13
re-run clang-format

FredKSchott:
I did a clean install of dependencies, and then ran clang-format. It looks like some formatting rules may have changed /shrug.

### DIFF
--- a/test/integration/browser.ts
+++ b/test/integration/browser.ts
@@ -301,9 +301,8 @@ function assertTestErrors(
         .to.have.members(
             Object.keys(actual),
             'Test file mismatch for ' + browser +
-                `: expected ${
-                              JSON.stringify(Object.keys(expected))
-                            } - got ${JSON.stringify(Object.keys(actual))}`);
+                `: expected ${JSON.stringify(Object.keys(expected))} - got ${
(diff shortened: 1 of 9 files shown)


test/integration/browser.ts
```diff
@@ -301,9 +301,8 @@ function assertTestErrors(
         .to.have.members(
             Object.keys(actual),
             'Test file mismatch for ' + browser +
-                `: expected ${
-                              JSON.stringify(Object.keys(expected))
-                            } - got ${JSON.stringify(Object.keys(actual))}`);
+                `: expected ${JSON.stringify(Object.keys(expected))} - got ${
+                    JSON.stringify(Object.keys(actual))}`);
```
usergenic: I actually agree with this clang-format update!

test/unit/context.ts
```diff
@@ -34,7 +34,7 @@ describe('Context', () => {
 
   describe('.plugins', () => {
 
-    it('excludes plugins with a falsy config', async() => {
+    it('excludes plugins with a falsy config', async () => {
```
usergenic: yay this!

PR #617, opened 2017-09-13
Fix broken npm scoped package support

FredKSchott:
Add packageName handling, that can handle NPM scoped packages. Do this instead of reading from the cwd.

### DIFF
--- a/runner/config.ts
+++ b/runner/config.ts
@@ -91,7 +91,7 @@ function readJsonSync(filename: string, dir?: string): any|null {
   try {
     return JSON.parse(config);
   } catch (e) {
-    console.error(`Could not parse ${configPath}`);
+    console.error(`Could not parse ${configPath} as JSON`);
     console.error(e);
   }
   return null;
--- a/runner/config.ts
+++ b/runner/config.ts
@@ -61,6 +61,7 @@ export interface Config {
     _servers?: {variant: string, url: string}[];
   };
   npm?: boolean;
+  packageName?: string;
   skipPlugins?: string[];
   sauce?: {};
   remote?: {};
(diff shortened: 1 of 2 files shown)


runner/config.ts
```diff
@@ -74,6 +75,50 @@ export interface Config {
...
+    return options.packageName;
+  }
+  const manifestName = (options.npm ? 'package.json' : 'bower.json');
+  const manifest = readJsonSync(manifestName, options.root);
+  if (manifest !== null) {
+    return manifest.name;
+  }
+  const basename = path.basename(options.root);
+  console.warn(
+      `no ${manifestName} found, defaulting to packageName=${basename}`);
```
usergenic: I like this

runner/config.ts
```diff
@@ -74,6 +75,50 @@ export interface Config {
...
+  let config: any;
+  try {
+    config = fs.readFileSync(configPath, 'utf-8');
+  } catch (e) {
+    return null;
+  }
+  try {
+    return JSON.parse(config);
+  } catch (e) {
+    console.error(`Could not parse ${configPath}`);
```
justinfagnani: maybe  add "as JSON"?

runner/webserver.ts
```diff
@@ -75,6 +76,12 @@ export function webserver(wct: Context): void {
       } catch (e) {
         // Safely ignore.
       }
+      const packageName = getPackageName(options);
+      const isPackageScoped = packageName && packageName[0] === '@';
+      if (isPackageScoped) {
+        browserScript = `../${browserScript}`;
```
justinfagnani: It'd be nice if we just calculated a "path to packages root" and then reused it as a prefix to address other package. This is fine for now.

PR #635, opened 2017-10-11
Convert browser-side code to typescript

rictic:
- [x] CHANGELOG.md not updated, this is a noop change

This brings it into the fold of maintainability.

I did it as tiny changes that separated out renaming from converting in the hopes that this would help github with diffing. It did not. Deliberately losing renaming metadata is one of git's original sins.


rictic:
(because the mini-commits don't seem to add a lot I'll probably squash this down to a single commit when merging)

PR #643, opened 2017-10-27
Forward proxy config

usergenic:
 - Supports passing proxy options to polyserve from the wct config.


rictic:
Tests are failing but it looks like a travis <-> sauce overload error, and this shouldn't cause problems with that binding.

usergenic:
I added some code to fix failures reported by `tsc` due to type strictness by either boxing them in with typeof checks (in the case of `typeof value === 'boolean' && value === false`) and added some `x: any` when `x` was implicitly any.  These occurred in `runner/config.ts` and `test/integration/browser.ts`.

Issue #523, opened 2017-04-04
WCT ignores the webserver hostname 


hvdb:
WCT uses hardcode localhost as the source for starting the tests.

https://github.com/Polymer/web-component-tester/blob/master/runner/webserver.ts#L207

If you use a selenium grid this is not working...
You can specify the webserver hostname but that is ignored here.

idoshamun:
What is the expected time to release this fix? It is very important for my project

PR #558, opened 2017-05-30
Use hostname instead of `localhost`

usergenic:
- [x] CHANGELOG.md has been updated
 - Fixes https://github.com/Polymer/web-component-tester/issues/523
 - BEFORE MERGING: Need to verify that WCT continues to work in situations where a call to `address().address` may return '::' for ipv6 scenarios.

PR #647, opened 2017-11-03
WCT uses webserver.hostname (#523)

kennethklee:
Replace hardcoded `localhost` to use the webserver hostname option.

This change allows users to utilise their own selenium grid as opposed to sauce labs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/).
 
 ## Unreleased
 <!-- Add new, unreleased items here. -->
+Fixed #523 WCT ignores the webserver hostname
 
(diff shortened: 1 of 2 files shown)


Bubbit:
Copy of #558 ?

kennethklee:
@Bubbit almost, except this is a simpler solution.

differences:
1. no changes to `httpbin.ts` since it's only used as a wct dev debug tool (which localhost would work fine with as it's run with `node runner/httpbin`)
2. relies on polyserve which already has the correct hostname instead of resolving a hostname from the server

kennethklee:
Really needing this. I'd prefer to use our own selenium grid rather than relying on saucelabs just for this.

justinfagnani:
Thanks @kennethklee! This will get in the next release

PR #657, opened 2017-12-07
Remove unnecessary overflow-y: auto

azakus:
This seems to be slowing down testing in Polymer and provides no
discernable visual difference in the test output


azakus:
Time to run full Polymer tests on Chrome 62 with `overflow-y: auto`: 444s
Time to run full Polymer tests on Chrome 62 _without_ `overflow-y: auto`: 323s

PR #666, opened 2018-01-11
Fix broken npm a11ySuite

FredKSchott:
## Background 

`a11ySuite` has become a module for npm/Polymer 3.0, so that it can import a necessary reference to Polymer. This has broken any consuming tests because modules don't block and execute after other traditional scripts (like the tests themselves). This code is annoyingly complicated and converting it all to async loading is nontrivial.

Instead of bending over backwards to maintain the current behavior in Polymer 3.0 and bringing a  working implicit reference to a window property (`window.a11ySuite`) into ESM from another ESM, it is actually simpler and a much healthier practice ti import this file directly in our tests (`import {a11ySuite} from 'wct-browser-legacy/a11ySuite.js`).

## What's Changed 

- a11ySuite now has an internal single source of truth: `/data/a11ySuite-source.js`
  - Used to generate the file for each environment (see next)
- `npm run build` generates two files: 
  - `+ /a11ySuite.js` is non-module, for bower & web-component-tester & non-npm usage
  - `+ /wct-browser-legacy/a11ySuite.js` is a module, for npm
- `generated_index.html`: 
  - References fixed to point to `/a11ySuite.js`
  - Moved from `/data/a11ySuite.js` (which now logs a deprecated warning message with info on new file location)
- `generated_index.html`:
  - No longer loads a11ySuite automatically for `--npm` use case
  - Modulizer will add import directly: 
    - `import {a11ySuite} from 'wct-browser-legacy/a11ySuite.js'`
  - Test writers will import it directly moving forward

---


data/index.html
```diff
@@ -13,9 +13,7 @@
   <% extraScripts.forEach(function(script) { %>
     <script src="<%- script %>"></script>
   <% }); %>
-  <% if (typeof npm !== 'undefined' && npm) { %>
-    <script type="module" src="../<%= a11ySuiteScript %>"></script>
-  <% } else { %>
+  <% if (typeof npm === 'undefined' || !npm) { %>
```
justinfagnani: is' this the same as just `!npm`?

wct-browser-legacy/a11ySuite.js
```diff
@@ -1,7 +1,8 @@
 import * as polymerDom from '../@polymer/polymer/lib/legacy/polymer.dom.js';
 const Polymer = { dom: polymerDom };
+export var a11ySuite;
```
justinfagnani: regenerate this file?

PR #686, opened 2018-03-09
Inject proper paths

### LINKED ISSUE #674: lodash: module is not defined
Hey!

I'm using web-component-tester just with a JS file. When testing I always get errors during test initialization.

image

Seems like something wrong with transpilation / index.js selection? Funnily enough, my actual tests still run. Its just that the end result never terminates with OK because that one thing fails.

e111077:

fixes #674 

In this PR we fix the issue where `browser.js` injects libraries to the wrong paths. An example where this happens:

```
my-element
 ├ test/
 | └ index.js
 └ node_modules/
   ├ lodash(v4)/
   └ wct-browser-legacy/
     ├ browser.js
     └ node_modules/
         └ lodash(v3)/
```

in this case `browser.js` wants to inject lodash v3, but `browser.js` simply injects `../../lodash/index.js` from `my-element/test/index.js`. this means that `browser.js` is actually pointing to lodash v4 which is not browser runnable. This PR makes it so that the runner:

1. Resolves `wct-browser-legacy` if in NPM mode
2. Resolves the injected library paths from `wct-browser-legacy`
   * or whatever npm package name you pass into `--wct-package-name`
3. Passes these paths to `browser.js` via `WCT.environmentScripts` in the generated index
4. Injects these resolved paths

So with these changes, the case above is now injecting `../../wct-browser-legacy/node_modules/lodash/index.js`. If you simply serve this file and run it without the runner, `browser.js` will simply inject what it did before. This works when there are no npm version conflicts.

You may want to look at [this commit](https://github.com/Polymer/web-component-tester/commit/4e50d3133fa2ed94ecd50c74c0db2f494cec9777) if you don't want to wade through all of the test fixtures I've added.

### DIFF
--- a/runner/config.ts
+++ b/runner/config.ts
@@ -17,6 +17,7 @@ import * as fs from 'fs';
 import * as _ from 'lodash';
 import * as nomnom from 'nomnom';
 import * as path from 'path';
+import * as resolve from 'resolve';
 import {Capabilities} from 'wd';
 
 import {BrowserDef} from './browserrunner';
@@ -43,7 +44,9 @@ export interface Config {
   testTimeout?: number;
   persistent?: boolean;
   extraScripts?: string[];
-  clientOptions?: {root?: string; verbose?: boolean};
+  wctPackageName?: string;
+  clientOptions?:
+      {root?: string; verbose?: boolean; environmentScripts?: string[]};
   activeBrowsers?: BrowserDef[];
   browserOptions?: {[name: string]: Capabilities};
   plugins?: (string|boolean)[]|{[key: string]: ({disabled: boolean} | boolean)};
@@ -82,6 +85,20 @@ export interface Config {
   };
   /** A deprecated option */
   browsers?: Browser[]|Browser;
+}
+
+export interface NPMPackage {
+  /**
+   * Name of the node package. e.g. '@polymer/polymer'
+   */
+  name: string;
+
+  /**
+   * JS entrypoints relative to packageName e.g. lodash/index.js would simply
+   * be ['index.js'] and myPackage/dist/addon.js and myPackage/lib/core.js would
+   * be ['dist/addon.js', 'lib/core.js']
+   */
+  jsEntrypoints: string[];
 }
 
 /**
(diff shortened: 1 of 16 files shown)


justinfagnani:
@e111077 you have some lint errors in here

runner/config.ts
```diff
@@ -128,6 +145,97 @@ export function getPackageName(options: Config): string|undefined {
...
+ * @param npmPackages List of NPMScript objects to be resolved
+ * @param wctPackageName Name of wct package with browser.js (Defaults to
+ * wct-browser-legacy)
+ */
+export function resolveWCTNPMEntrypointNames(
+    config: Config, npmPackages: NPMPackage[],
+    wctPackageName = 'wct-browser-legacy'): string[] {
+  // grab from CLI flag defaults to wct-browser-legacy
+  wctPackageName =
+      config.wctPackageName ? config.wctPackageName : wctPackageName;
```
justinfagnani: do explicit null or undefined checks, rather than rely on implicit ToBoolean conversion, so `config.wctPackageName === undefined`

justinfagnani: It's interesting that the config overrides the parameter. I would think the parameter if specified would override. When would the parameter be passed? Instead of this precedence, consider:

```ts
wctPackageName?): string[] {
  if (wctPackageName === undefined) {
    wctPackageName = config.wctPackageName;
  }
}
```

runner/config.ts
```diff
@@ -128,6 +145,97 @@ export function getPackageName(options: Config): string|undefined {
...
+    absoluteBrowserPath = resolve.sync(wctPackageName, {basedir: config.root});
+  } catch {
+    throw new Error(
+        `${wctPackageName} not installed. Please change --wct-package-name` +
+        ` flag or install the package.`);
+  }
+
+  // We want to find and inject dependencies WCT relies on not the local
+  // package or its dependencies' dependencies
+  const absoluteWCTRoot =
```
justinfagnani: the async version of `resolve()` has a `pathFilter` option that might let you get the package dir without the `truncatePathToDir` function: https://github.com/browserify/resolve#resolveid-opts-cb

I think you can return the empty string and get what you want, and you can use the async version because `resolveWCTNPMEntrypointNames` is called from an async function.

runner/config.ts
```diff
@@ -128,6 +145,97 @@ export function getPackageName(options: Config): string|undefined {
...
+      truncatePathToDir(wctPackageName, absoluteBrowserPath);
+  const resolvedEntrypoints: string[] = [];
+
+  for (const npmPackage of npmPackages) {
+    const absoluteNPMMainPath =
```
justinfagnani: style nit: only capitalize the first letter acronyms of 3 or longer: `absoluteNpmMainPath`

runner/webserver.ts
```diff
@@ -36,6 +36,22 @@ const DEFAULT_HEADERS = {
   'Expires': '0',
 };
 
+// scripts to be injected into the running test
+const ENVIRONMENT_SCRIPTS: NPMPackage[] = [
+  {name: 'stacky', jsEntrypoints: ['browser.js']},
```
justinfagnani: do we have cases with more than one entrypoint?

runner/config.ts
```diff
@@ -128,6 +145,97 @@ export function getPackageName(options: Config): string|undefined {
...
+ * Finds the first npm scope starting with `@` from a path. Undefined if no path
+ * found
+ *
+ * @param path Path to be searched
+ */
+export function getPackageScopeFromPath(path: string): string|undefined {
+  const pathParts = path.split('/');
+
+  for (const pathPart of pathParts) {
+    if (pathPart[0] === '@') {
```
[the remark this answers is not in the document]
e111077: actually I no longer use this method. Also, I took this algo from [here](https://github.com/Polymer/web-component-tester/blob/master/runner/webserver.ts#L80)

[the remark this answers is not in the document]
justinfagnani: The difference is that's checking the first character of the package name, which has already been determined by `getPackageName()`. Here you're iterating over all the parts in a path. There's nothing stating that a path part might not be the start of the package name, but still start with a '@'.

runner/config.ts
e111077: I can't seem to get `pathFilter` to even get called (a me problem probs). Regardless, it seems as if `resolve()` returns undefined rather than a promise.

runner/webserver.ts
e111077: We do not. Do you think it's better to change` jsEntrypoints: string[]` to` jsEntrypoint: string`?

justinfagnani: I think it's simpler, if we're not using multiple entrypoints yet. `resolve()` works with full file paths, so you wouldn't need an object here, just strings.

runner/config.ts
```diff
@@ -128,6 +144,82 @@ export function getPackageName(options: Config): string|undefined {
...
 }
 
+/**
+ * Truncates the path to the slash after the last occurrence of the given
+ * package name.
+ *
+ * @param directory Name of directory
+ * @param path Path to be truncated
+ */
+export function truncatePathToDir(directory: string, path: string): string|
```
FredKSchott: we end up reimplementing these sorts of path manipulation helpers in so many of our different libraries. It's at the point where we should start looking into pulling them into a shared helper library.

FredKSchott: don't re-declare, path is already an imported module

runner/config.ts
```diff
@@ -17,6 +17,7 @@ import * as fs from 'fs';
 import * as _ from 'lodash';
 import * as nomnom from 'nomnom';
 import * as path from 'path';
```
FredKSchott: something to keep in mind: path's behavior changes based on the machine it's running on (ex: outputs/expects `/` for Unix & `\` for windows). I see you using it in a few places below, consider using `path.posix` if you want to force URL-compliant slashes on all machines.

https://nodejs.org/api/path.html#path_path_posix
ex: `path.posix.relative(...`

runner/webserver.ts
```diff
@@ -78,6 +94,19 @@ export function webserver(wct: Context): void {
       }
       const packageName = getPackageName(options);
       const isPackageScoped = packageName && packageName[0] === '@';
+
+      // concat options.clientOptions.environmentScripts with resolved
+      // ENVIRONMENT_SCRIPTS
+      options.clientOptions =
```
FredKSchott: shorten to: 
```
options.clientOptions = options.clientOptions || {};
options.clientOptions.environmentScripts = options.clientOptions.environmentScripts || {}
etc.
```

runner/config.ts
e111077: great catch

runner/config.ts
e111077: Trying this again and I simply cannot get `pathFilter` to work. Running this in node from the `test/fixtures/fake-packages/duplicated-dep` directory

```js
resolve('dependency', {
  pathFilter: function(pkg, path, relPath) {
    console.log('PFILTER ARGS:',arguments);
    return path;
  }
},
function() { console.log(`CB ARGS:`, arguments) });
```

it just wont call, and we're using the latest release of `resolve`

runner/config.ts
e111077: Any idea on a good place to do that?

justinfagnani: Oh, for a monorepo... polymer-analyzer, tools-common and polymer-project-config are the only really common packages we have right now. None of them seem great.

PR #691, opened 2018-03-21
prepare changelog and update polyserve dependency

FredKSchott:


FredKSchott:
verbal +1 from @justinfagnani

PR #697, opened 2018-04-04
Handle content loading timeouts more gracefuly.

rictic:
This converts some errors that look like `Test run ended in failure: Error thrown outside of test function: Cannot read property 'WCT' of null`

into ones that look like `Timed out loading http://localhost/:17604/components/my-component/test.html`

 - [x] CHANGELOG.md not updated, minor internal change


rictic:
WCT's tests are broken at master. This change seems very safe, changing one error message for another, better one. Merging despite red.

PR #710, opened 2018-04-12
Updated versions of browser.js and add findup-sync back to dependencies

usergenic:
Running `npm run build` produced different versions of browser.js than what was in the repo.  I think someone didn't commit changes to these generated files when they changed the `.ts` source.  Added findup-sync back to dependencies because was used by browser.js


usergenic:
After this PR is merged I will release wct-browser-legacy and then we probably should release WCT again.

aomarks:
Why are these checked in rather than just part of the publish step?

usergenic:
Okay this PR includes the added-back findup-sync library.

PR #726, opened 2018-07-17
Clear share from ChildRunner to prevent memory leak. Fixes #579

### LINKED ISSUE #579: WCT.loadSuites memory leak
bumping https://github.com/Polymer/web-component-tester/issues/546 with more information

browser.js is creating a new iFrame for every test suite.  This creates a new window object which will not get GC'd.  You end up with all the globals stuck in memory for every test suite.  Changing the implementation to reuse the same iFrame vs creating a new one does not fix the issue.

Simple example.  If I had 4 tests in a test suite run that all set window.foo = 123.  You will see window.foo 4 times in the heap at the end of the test run.  This is much worse in a real example because of how much is imported.

Expected:  each test suite runs with a clean memory footprint
Actual:  each test suite runs with memory leaks from the previous test suite runs

kevinpschaaf:
Ensure that the `share` property pulled out of the child iframe is cleared from the `ChildRunner` instance when the iframe is cleaned up, to prevent retaining the iframe and everything inside of it.

There's not a great unit test to add, since this addresses a memory leak rather than a functional error.

Fixes #579.


azakus:
@usergenic where does the source code for this live now?

kevinpschaaf:
@azakus @rictic Any clue about the status of tests over on this repo? Master seems pretty broken?

usergenic:
https://github.com/Polymer/tools is the new source of truth and where we cut releases from.  I've not been tracking whether polymer 2.0 bower-supporting fixes to wct have needed backporting into here, but I'll make a note to do that if these tests arent passing etc.

usergenic (approved):

